PR #2, opened 2022-12-16
Relax version requirement for Qulacs

kwkbtr:
Relaxed the version specification for Qulacs in `quri-parts-qulacs` from `^0.3.0` to `>=0.3.0, <0.6.0` so that the latest version 0.5.3 can be installed.


r-imai-67 (approved):
LGTM!

PR #10, opened 2022-12-26
estimator fix for pauli Identity group

Kushargs:
The estimator has an implicit assumption about the Pauli Identity being in a separate group. This has remained consistent over the current bitwise implementation. A grouping that puts the Pauli Identity into a different set is valid and possible. The current implementation of the estimator will incorrectly calculate the Expectation value for the aforementioned case.

The recommended fix was to set `const` to zero whenever the Pauli Identity is not in the standalone group. This PR captures the same.

### DIFF
--- a/packages/core/quri_parts/core/estimator/sampling/estimator.py
+++ b/packages/core/quri_parts/core/estimator/sampling/estimator.py
@@ -121,8 +121,16 @@ def sampling_estimate(
         if len(op) == 1:
             return _ConstEstimate(const)
 
-    measurements = measurement_factory(op)
-    measurements = [m for m in measurements if m.pauli_set != {PAULI_IDENTITY}]
+    # If there is a standalone Identity group then eliminate, else set const 0.
+    standalone = False
+    measurements = []
+    for m in measurement_factory(op):
+        if m.pauli_set == {PAULI_IDENTITY}:
+            standalone = True
+        else:
+            measurements.append(m)
+    if not standalone:
+        const = 0
 
     pauli_sets = tuple(m.pauli_set for m in measurements)
     shot_allocs = shots_allocator(op, pauli_sets, total_shots)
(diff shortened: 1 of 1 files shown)


Kushargs:

Kushargs:
recheck

packages/core/quri_parts/core/estimator/sampling/estimator.py
```diff
@@ -121,8 +121,16 @@ def sampling_estimate(
...
+    # If there is a standalone Identity group then eliminate, else set const 0.
+    standalone = False
+    measurements = []
+    for m in measurement_factory(op):
+        if m.pauli_set == {PAULI_IDENTITY}:
+            standalone = True
+        else:
+            measurements.append(m)
+    if not standalone:
+        const = 0
```
kwkbtr: Combined with the above code block, it can be refactored as follows, what do you think?

```suggestion
    if len(op) == 1 and PAULI_IDENTITY in op:
        return _ConstEstimate(op[PAULI_IDENTITY])
          
    # If there is a standalone Identity group then eliminate, else set const 0.
    const: complex = 0.0
    measurements = []
    for m in measurement_factory(op):
        if m.pauli_set == {PAULI_IDENTITY}:
            const = op[PAULI_IDENTITY]
        else:
            measurements.append(m)
```

kwkbtr (commented):
Thank you for the fix!

packages/core/quri_parts/core/estimator/sampling/estimator.py
Kushargs: Yes, this looks better to me.

Kushargs: This would have duplicated the operator being the `PAULI_IDENTITY` condition. So, I have added a new commit.

kwkbtr (approved):
Thanks!

PR #11, opened 2022-12-28
Remodulation of `braket.ht` module as subpart of `braket.backend` 

Zshan0:
The original `braket.ht` module had two functionalities, namely `graph_state_circuit` and `device_connectivity_graph`. These two functionalities are tackled as follows:

- `graph_state_circuit` was a function that constructed the required circuit for HT in the `braket` framework. This can be removed as the circuit can be implemented in native `qp` and then converted to `braket` if necessary. The required circuit has already been made available in the `ht` module.
- `device_connectivity_graph` is a backend-specific functionality that can be moved to `braket.backend` as it is independent of the process that demands the same.


kwkbtr (approved):
LGTM!

PR #15, opened 2023-01-20
Corrected arguments of add gate methods

lqtmirage:
- For the following gates, the argument names of add_[gate name]_gate() methods of circuit were changed to match the argument names of gate objects
  - U1 / U2 / U3 / Pauli / PauliRotation
- Prepare add_Identity_gate() method
- Update related tests


kwkbtr (approved):
👍

PR #18, opened 2023-01-24
IonQ native gates transpiler

lqtmirage:
- Add ionq package
- Unique native gates for IonQ: GPi, GPi2, and MS
- Transpiler to decompose into IonQ native gates

*Note*

- The conversion process to the native gates is a direct porting of the process of the code in IonQ's introductory article.
  - https://ionq.com/docs/getting-started-with-native-gates#introducing-the-native-gates
- Due to the nature of the conversion process, each qubit may acquire an arbitrary extra phase. Therefore, only the measurement in the computational basis is guaranteed to match between before and after conversion.


kwkbtr (approved):
👍

PR #19, opened 2023-01-25
Numerical differentiation

toru4838:
Add functions that returns the {gradient / hessian} of a given function at it's parameters.

For gradient and hessian, users can select from
- forward difference formula
- backward difference formula
- central difference formula.

`gradient()` and `hessian()` are the aliases of `central_difference_gradient_formula()` and `central_difference_hessian_formula()`


kwkbtr (approved):
👍

PR #28, opened 2023-02-20
Toffoli to 2 qubit gates transpiler

lqtmirage:
- Add a transpiler to decompose Toffoli gates into 2 qubit gate sequences
- Add the Toffoli decomposer to pipelines for several gate sets


kwkbtr (approved):
👍

PR #33, opened 2023-03-06
Add ITensor simulator backend

speed1313:
# Summary

- ITensor MPS simulator is supported as a backend
- Estimator and Sampler is supported


speed1313:

packages/circuit/tests/circuit/test_parameter_mapping.py
```diff
@@ -52,7 +52,7 @@ def m(vals: ParameterValueAssignment) -> float:
 
                 return m
 
-            def is_trivial_mapping(self) -> bool:
+            def is_trivial_mapping(self) -> bool:  # type: ignore[override]
```
toru4838: This will fix the error
```suggestion
            @property
            def is_trivial_mapping(self) -> bool:
```

packages/itensor/quri_parts/itensor/estimator.py
```diff
@@ -0,0 +1,204 @@
...
+
+abs_dir = os.path.dirname(os.path.abspath(__file__))
+library_path = os.path.join(abs_dir, "library.jl")
+include_statement = 'include("' + library_path + '")'
+jl.seval(include_statement)
+
+
+class _Estimate(NamedTuple):
+    value: complex
+    error: float = 0.0
```
toru4838: I'm not sure if ITensor is an exact simulator. Usually TN calculation involves approximation when creating MPS. If that is the case in ITensor as well , `error=0.0` might cause misunderstanding.
You can set `None` if it's not exact.

packages/itensor/tests/itensor/circuit/test_itensor_convert_circuit.py
```diff
@@ -0,0 +1,164 @@
...
+def test_convert_three_qubit_gate() -> None:
+    qubits = 3
+    s: juliacall.VectorValue = jl.siteinds("Qubit", qubits)
+    psi: juliacall.AnyValue = jl.initState(s, qubits)
+    expected_list = [
+        [1.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0],
+    ]
+    for i, qp_fac in enumerate(three_qubit_gate_list):
+        circuit = QuantumCircuit(qubits)
+        circuit.add_gate(qp_fac(0, 1, 2))
```
toru4838: Maybe |110> or |111> ?

packages/itensor/tests/itensor/operator/test_itensor_convert_operator.py
```diff
@@ -0,0 +1,25 @@
...
+jl.seval("using ITensors")
+
+
+def test_convert_operator() -> None:
+    s: juliacall.VectorValue = jl.siteinds("Qubit", 6)
+    pauli = pauli_label("Z0 Z2 Z5")
+    itensor_op = convert_operator(pauli, s)
+    psi: juliacall.AnyValue = jl.initState(s)
+    exp: float = jl.expectation(psi, itensor_op)
+    assert exp == -1.0
```
toru4838: This is fine. Please add a test for `Operator` also 👍

toru4838 (commented):
Thanks for PR. Great work!

packages/itensor/tests/itensor/circuit/test_itensor_convert_circuit.py
```diff
@@ -0,0 +1,164 @@
...
+    s: juliacall.VectorValue = jl.siteinds("Qubit", qubits)
+    psi: juliacall.AnyValue = jl.initState(s, qubits)
+    expected_list = [
+        [1.0, 0.0, 0.0, 0.0],
+        [1.0, 0.0, 0.0, 0.0],
+        [1.0, 0.0, 0.0, 0.0],
+    ]
+    for i, qp_fac in enumerate(two_qubit_gate_list):
+        circuit = QuantumCircuit(qubits)
+        circuit.add_gate(qp_fac(0, 1))
```
toru4838: This can be better if the initial state is |10>

speed1313:
Thank you for your review!
I have fixed it.

packages/itensor/quri_parts/itensor/estimator.py
```diff
@@ -0,0 +1,204 @@
...
+
+abs_dir = os.path.dirname(os.path.abspath(__file__))
+library_path = os.path.join(abs_dir, "library.jl")
+include_statement = 'include("' + library_path + '")'
+jl.seval(include_statement)
+
+
+class _Estimate(NamedTuple):
+    value: complex
+    error: Union[float, None] = None
```
toru4838: Sorry for the confusion but now I think this is better
```suggestion
class _Estimate(NamedTuple):
    value: complex
    error: float = np.nan
```

You can follow the implementation
https://github.com/QunaSys/quri-parts/blob/c335623bce3a498b7c9a6e416ca8abbc95557317/packages/algo/quri_parts/algo/mitigation/cdr/cdr.py#L39-L41

toru4838 (approved):
LGTM!
Let's wait for @kwkbtr to review

packages/itensor/quri_parts/itensor/estimator.py
```diff
@@ -0,0 +1,205 @@
...
+) -> Sequence[Estimate[complex]]:
+    qubits = state.qubit_count
+    s: juliacall.VectorValue = jl.siteinds("Qubit", qubits)
+    psi: juliacall.AnyValue = jl.initState(s, qubits)
+    circuit = convert_circuit(state.circuit, s)
+    psi = jl.apply(circuit, psi)
+    results = []
+    for op in operators:
+        itensor_op = convert_operator(op, s)
+        results.append(_Estimate(value=jl.expectation(psi, itensor_op)))
```
toru4838: This part causes error when there is an empty `Operator` in `operators`.



```python
concurrent_mps_estimator([Operator({...}), {}, …], [state])
```

   → JuliaError: OpSum has no terms
```suggestion
    for op in operators:
        if op == zero():
            results.append(_Estimate(value=0.0))
            continue
        itensor_op = convert_operator(op, s)
        results.append(_Estimate(value=jl.expectation(psi, itensor_op)))
```

pyproject.toml
kwkbtr, line 1: Were these blank lines removed automatically by Poetry?

packages/itensor/quri_parts/itensor/library.jl
```diff
@@ -0,0 +1,142 @@
...
+function add_three_qubit_gate(gate_list::Vector, gate_name::String, control_index1::Integer, control_index2::Integer, target_index::Integer)::Vector
+    push!(gate_list, (gate_name, control_index1, control_index2, target_index))
+    return gate_list
+end
+
+function add_single_qubit_rotation_gate(gate_list::Vector, gate_name::String, target_index::Integer, param::Number)::Vector
+    if gate_name == "Rx" || gate_name == "Ry"
+        push!(gate_list, (gate_name, target_index, (θ=param,)))
+    elseif gate_name == "Rz"
+        push!(gate_list, (gate_name, target_index, (ϕ=param,)))
```
kwkbtr: Is `θ` better here, since `ϕ` seems to be deprecated?

https://github.com/ITensor/ITensors.jl/blob/5afe9e84f90c6f0d116797d4f6c10a4a0de997f7/src/physics/site_types/qubit.jl#L178-L187

packages/itensor/quri_parts/itensor/operator/__init__.py
```diff
@@ -0,0 +1,30 @@
...
+from juliacall import Main as jl
+
+from quri_parts.core.operator import Operator, PauliLabel, pauli_name
+
+
+def convert_operator(
+    operator: Union[Operator, PauliLabel], s: juliacall.VectorValue
+) -> juliacall.AnyValue:
+    """Convert an :class:`~Operator` or a :class:`~PauliLabel` to an ITensor
+    MPO operator."""
```
kwkbtr: It is unclear what `s` means here.
I prefer more readable argument name and docstring.

packages/itensor/quri_parts/itensor/operator/__init__.py
```diff
@@ -0,0 +1,30 @@
...
+        paulis = [(operator, 1)]
+    os: juliacall.AnyValue = jl.OpSum()
+    for pauli, coef in paulis:
+        pauli_gates: juliacall.VectorValue = jl.gate_list()
+        if len(pauli) == 0:
+            os = jl.add_coef_identity(os, coef, s)
+            continue
+        for i, p in pauli:
+            pauli_gates = jl.add_pauli(pauli_gates, pauli_name(p), i + 1)
+        os = jl.add_coef_pauli(os, coef, pauli_gates)
```
kwkbtr: About `pauli_gates` naming: the `PauliLabel`s handled here are not considered as gates, so I prefer avoiding "gates" in variable names.

packages/itensor/quri_parts/itensor/library.jl
```diff
@@ -0,0 +1,142 @@
...
+    return pauli_gates
+end
+
+function add_coef_pauli(os, coefficient::Number, pauli_gates)
+    os += (coefficient, pauli_gates...)
+    return os
+end
+
+function add_coef_identity(os, coefficient::Number, s)
+    os += (coefficient,[isodd(i) ? "I" : div(i,2) for i in 1:(2* length(s))]...)
```
kwkbtr: I'm not familiar with ITensor but I wonder if this array (`["I", 1, "I", 2, ...]`), i.e. identity operators on all sites, is necessary.
I think in theory an identity operator on only one site (i.e. `["I", 1]`) or even empty (`[]`?) are equivalent, though I'm not sure they are acceptable by ITensor.

packages/itensor/quri_parts/itensor/library.jl
```diff
@@ -0,0 +1,142 @@
...
+function add_pauli(pauli_gates::Vector, pauli_name::String, target_index::Integer)::Vector
+    push!(pauli_gates, pauli_name)
+    push!(pauli_gates, target_index)
+    return pauli_gates
+end
+
+function add_coef_pauli(os, coefficient::Number, pauli_gates)
+    os += (coefficient, pauli_gates...)
+    return os
+end
```
kwkbtr: So `add_pauli` *multiplies* a Pauli operator *on a single site* to *a Pauli product*, whereas `add_coef_pauli` adds *a Pauli product* to an operator.
Could you consider less confusing names for these functions?

packages/itensor/quri_parts/itensor/circuit/__init__.py
```diff
@@ -0,0 +1,123 @@
...
+    gate_names.Z: "Z",
+    gate_names.H: "H",
+    gate_names.S: "S",
+    gate_names.Sdag: "Sdag",
+    gate_names.SqrtX: "√X",
+    gate_names.SqrtXdag: "SqrtXdag",
+    gate_names.SqrtY: "SqrtY",
+    gate_names.SqrtYdag: "SqrtYdag",
+    gate_names.T: "T",
+    gate_names.Tdag: "Tdag",
```
kwkbtr: Since `SqrtX=√X`, we may be able to use `√Xdag`, `√Y` and `√Ydag`?
(`dag=†` may also be possible, but it may look too similar to `+` and confusing ;)

packages/itensor/quri_parts/itensor/circuit/__init__.py
```diff
@@ -0,0 +1,123 @@
...
+}
+
+_three_qubit_gate_itensor: Mapping[ThreeQubitGateNameType, str] = {
+    gate_names.TOFFOLI: "Toffoli",
+}
+
+
+def convert_circuit(
+    circuit: NonParametricQuantumCircuit, s: juliacall.VectorValue
+) -> juliacall.VectorValue:
```
kwkbtr: It is unclear what `s` means here.
I prefer more readable argument name and docstring.

packages/itensor/quri_parts/itensor/sampler.py
```diff
@@ -0,0 +1,71 @@
...
+library_path = os.path.join(abs_dir, "library.jl")
+include_statement = 'include("' + library_path + '")'
+jl.seval(include_statement)
+
+
+def _sample(circuit: NonParametricQuantumCircuit, shots: int) -> MeasurementCounts:
+    qubits = circuit.qubit_count
+    s: juliacall.VectorValue = jl.siteinds("Qubit", qubits)
+    psi: juliacall.AnyValue = jl.initState(s, qubits)
+    qs_circuit = convert_circuit(circuit, s)
```
kwkbtr: Minor comment: the name `qs_circuit` doesn't look like a meaningful name; for example, since `convert_circuit` converts the circuit into a sequence of `op`s, we can call it like `circuit_ops`?

packages/itensor/quri_parts/itensor/library.jl
```diff
@@ -0,0 +1,142 @@
...
+    (-1+im)/2 (1-im)/2
+]
+ITensors.op(::OpName"Tdag", ::SiteType"Qubit") = [
+    1 0
+    0 1/sqrt(2)-im/sqrt(2)
+]
+
+
+
+function initState(s, qubits::Integer)
```
kwkbtr: Since all other function names are in snake case, I think this function should be `init_state()`.

packages/itensor/quri_parts/itensor/sampler.py
```diff
@@ -0,0 +1,71 @@
...
+    psi = jl.apply(qs_circuit, psi)
+    result: list[int] = []
```
kwkbtr: Is this line necessary?

packages/itensor/quri_parts/itensor/sampler.py
```diff
@@ -0,0 +1,71 @@
...
+    concurrency: int = 1,
+) -> Iterable[MeasurementCounts]:
+    return execute_concurrently(
+        _sample_sequentially, None, circuit_shots_tuples, executor, concurrency
+    )
+
+
+# For now, this function works when the executor is defined like below
+# `with ProcessPoolExecutor(max_workers=2, mp_context=get_context("spawn"))
+# as executor:`
```
kwkbtr: This information should be known to the user of this function, so let's move this inside the docstring.

packages/itensor/quri_parts/itensor/estimator.py
```diff
@@ -0,0 +1,205 @@
...
+from .circuit import convert_circuit
+from .operator import convert_operator
+
+if TYPE_CHECKING:
+    from concurrent.futures import Executor
+
+abs_dir = os.path.dirname(os.path.abspath(__file__))
+library_path = os.path.join(abs_dir, "library.jl")
+include_statement = 'include("' + library_path + '")'
+jl.seval(include_statement)
```
kwkbtr: - Exactly the same lines are also in `sampler.py`, so I think it is better to put them into a function.
- These lines are executed just after this module is imported, so it impacts start-up time of the program, even if functions using ITensor are not executed at all. Is it possible to delay this initialization to just before executing such functions? For example, the initialization function can be named as `ensure_itensor_loaded()` and each function using ITensor can call it at the beginning.

packages/itensor/quri_parts/itensor/estimator.py
```diff
@@ -0,0 +1,205 @@
...
+    circuit = convert_circuit(state.circuit, s)
+
+    # create ITensor operator
+    op = convert_operator(operator, s)
+
+    # calculate expectation value
+    psi = jl.apply(circuit, psi)
+    exp: float = jl.expectation(psi, op)
+
+    return _Estimate(value=exp)
```
kwkbtr: I'm not 100% confident, but isn't the estimation error zero (`error=0.0`) for this case?
If the expectation value is calculated by inner product, there should be no statistical error.

packages/itensor/quri_parts/itensor/estimator.py
```diff
@@ -0,0 +1,205 @@
...
+        if num_ops == 1:
+            operators = [next(iter(operators))] * num_states
+        return execute_concurrently(
+            _sequential_estimate, None, zip(operators, states), executor, concurrency
+        )
+
+
+# For now, this function works when the executor is defined like below
+# `with ProcessPoolExecutor(max_workers=2, mp_context=get_context("spawn"))
+# as executor:`
```
kwkbtr: This information should be known to the user of this function, so let's move this inside the docstring.

packages/itensor/quri_parts/itensor/estimator.py
```diff
@@ -0,0 +1,205 @@
...
+def create_itensor_mps_parametric_estimator() -> ParametricQuantumEstimator[
+    ITensorParametricStateT
+]:
+    def estimator(
+        operator: Estimatable, state: ITensorParametricStateT, param: Sequence[float]
+    ) -> Estimate[complex]:
+        ests = _sequential_parametric_estimate((operator, state), [param])
+        return ests[0]
+
+    return estimator
```
kwkbtr: I guess you can write like:
```suggestion
    return create_parametric_estimator(create_itensor_mps_estimator())
```

packages/itensor/tests/itensor/circuit/test_library.jl
kwkbtr, line 1: I thought this is a test file for `library.jl` and got a bit confused...

packages/itensor/tests/itensor/circuit/test_library.jl
```diff
@@ -0,0 +1,15 @@
+using ITensors
+function initState(s, qubits::Integer)
+    psi = productMPS(s, ["0" for i in 1:qubits])
+    return psi
+end
```
kwkbtr: Is this different from `initState()` in `library.jl`?

packages/itensor/tests/itensor/circuit/test_itensor_convert_circuit.py
kwkbtr, line 1: I think applying the circuit and comparing the state vector is okay, but it would be more straightforward if we can directly compare the converted sequence of operators.

packages/itensor/tests/itensor/estimator/test_itensor_estimator.py
```diff
@@ -0,0 +1,180 @@
...
+            ComputationalBasisState(6, bits=0b110000),
+            ComputationalBasisState(6, bits=0b110010),
+        ]
+        with ProcessPoolExecutor(
+            max_workers=2, mp_context=get_context("spawn")
+        ) as executor:
+            estimator = create_itensor_mps_concurrent_estimator(executor, concurrency=2)
+            result = estimator(operators, states)
+        assert result[0].value == -1  # type: ignore
+        assert result[1].value == -0.25 + 0.5j  # type: ignore
```
kwkbtr: Are these `# type: ignore` necessary?

packages/itensor/tests/itensor/operator/test_itensor_convert_operator.py
kwkbtr, line 1: I think comparing expectation values is okay, but it would be more straightforward if we can directly compare the converted operators.

packages/itensor/tests/itensor/estimator/test_itensor_estimator.py
```diff
@@ -0,0 +1,180 @@
...
+                pauli_label("Z1 Z2 Z4"): 0.5j,
+            }
+        )
+        state = ComputationalBasisState(6, bits=0b110010)
+        estimator = create_itensor_mps_estimator()
+        estimate = estimator(operator, state)
+        assert estimate.value == -0.25 + 0.5j
+
+
+class TestITensorConcurrentEstimator:
```
kwkbtr, line 42: Each test case in this group takes about 1 minutes to finish, which I think a bit too slow...

```
56.59s call     packages/itensor/tests/itensor/estimator/test_itensor_estimator.py::TestITensorConcurrentEstimator::test_concurrent_estimate
56.42s call     packages/itensor/tests/itensor/estimator/test_itensor_estimator.py::TestITensorConcurrentEstimator::test_concurrent_estimate_single_operator
55.16s call     packages/itensor/tests/itensor/estimator/test_itensor_estimator.py::TestITensorConcurrentEstimator::test_concurrent_estimate_single_state
```

packages/itensor/tests/itensor/sampler/test_itensor_sampler.py
```diff
@@ -0,0 +1,53 @@
...
+        sampler = create_itensor_mps_sampler()
+        counts = sampler(circuit, shots)
+
+        assert set(counts.keys()).issubset(range(2**qubits))
+        assert all(c >= 0 for c in counts.values())
+        assert sum(counts.values()) == shots
+
+
+class TestITensorMPSConcurrentSampler:
+    def test_concurrent_sampler(self) -> None:
```
kwkbtr, line 40: This test case is a bit slow too...

```
47.60s call     packages/itensor/tests/itensor/sampler/test_itensor_sampler.py::TestITensorMPSConcurrentSampler::test_concurrent_sampler
```

packages/itensor/tests/itensor/estimator/test_itensor_estimator.py
kwkbtr: I took profiling of the slow test cases but only found that `{method 'acquire' of '_thread.lock' objects}` spends all the time...
Until we figure out the reason it may be better to skip those test cases involving concurrent execution with `ProcessPoolExecutor`.

image

packages/itensor/quri_parts/itensor/estimator.py
speed1313: Oh! Thank you!

pyproject.toml
speed1313: No, that's because I edited it manually. 
I'll reinsert the blanks.

packages/itensor/quri_parts/itensor/library.jl
speed1313: I didn't know that. Thank you!

packages/itensor/quri_parts/itensor/library.jl
speed1313: I tried it with ["I", 1] and was able to do it.
Therefore, change it to ["I", 1].
Thank you.

packages/itensor/quri_parts/itensor/circuit/__init__.py
speed1313: Sorry, I don't know what you mean. 
Could you please tell me more about it?

kwkbtr: Maybe I should have commented on `library.jl`.
For example in `library.jl` `SqrtY` is defined as:

```julia
ITensors.op(::OpName"SqrtY", ::SiteType"Qubit") = [
    (1+im)/2 (-1-im)/2
    (1+im)/2 (1+im)/2
]
```

However, since `√X` is defined in ITensor, we may be able to follow the same convention and use `√Y` as the gate name instead of `SqrtY`.

speed1313: I see, thank you!

packages/itensor/quri_parts/itensor/estimator.py
speed1313: > If the expectation value is calculated by inner product, there should be no statistical error.

Yes, but mps simulator calculates the inner product in an approximate manner, which introduces approximate error due to approximation. 
If error = 0.0 though approximation error is happend, user may be confused.
How should this error be expressed?

kwkbtr: I agree that the error coming from the approximation should be taken into consideration.
Does current implementation do approximation?
I didn't find configuration for approximation (e.g. low rank approx.), so I assumed that for now no approximation is used.

packages/itensor/tests/itensor/circuit/test_library.jl
speed1313: No, I'll delete it.

packages/itensor/tests/itensor/estimator/test_itensor_estimator.py
speed1313: No, I'll delete it, thank you.

packages/itensor/quri_parts/itensor/library.jl
```diff
@@ -81,7 +80,7 @@ function add_single_qubit_rotation_gate(gate_list::Vector, gate_name::String, ta
     if gate_name == "Rx" || gate_name == "Ry"
         push!(gate_list, (gate_name, target_index, (θ=param,)))
     elseif gate_name == "Rz"
-        push!(gate_list, (gate_name, target_index, (ϕ=param,)))
+        push!(gate_list, (gate_name, target_index, (θ=param,)))
```
kwkbtr: Now these two branches can be unified.

packages/itensor/quri_parts/itensor/estimator.py
```diff
@@ -47,11 +42,12 @@ class _Estimate(NamedTuple):
 
 
 def _estimate(operator: Estimatable, state: ITensorStateT) -> Estimate[complex]:
+    ensure_itensor_loaded(__file__)
     if operator == zero():
         return _Estimate(value=0.0)
```
kwkbtr:
```suggestion
        return _Estimate(value=0.0, error=0.0)
```

packages/itensor/quri_parts/itensor/estimator.py
```diff
@@ -82,13 +78,17 @@ def _sequential_estimate(
...
     qubits = state.qubit_count
     s: juliacall.VectorValue = jl.siteinds("Qubit", qubits)
-    psi: juliacall.AnyValue = jl.initState(s, qubits)
+    psi: juliacall.AnyValue = jl.init_state(s, qubits)
     circuit = convert_circuit(state.circuit, s)
     psi = jl.apply(circuit, psi)
     results = []
     for op in operators:
+        if op == zero():
+            results.append(_Estimate(value=0.0))
```
kwkbtr:
```suggestion
            results.append(_Estimate(value=0.0, error=0.0))
```

packages/itensor/quri_parts/itensor/estimator.py
```diff
@@ -82,13 +78,17 @@ def _sequential_estimate(
...
+            continue
         itensor_op = convert_operator(op, s)
         results.append(_Estimate(value=jl.expectation(psi, itensor_op)))
```
kwkbtr:
```suggestion
        results.append(_Estimate(value=jl.expectation(psi, itensor_op), error=0.0))
```

packages/itensor/quri_parts/itensor/load_itensor.py
```diff
@@ -0,0 +1,10 @@
+import os
+
+from juliacall import Main as jl
+
+
+def ensure_itensor_loaded(calling_file_path: str) -> None:
```
kwkbtr: Rather than receiving `calling_file_path`, I think it is better to use the file name of this file.
If `__file__` is unavailable in the function, maybe you can capture it just above the function like `_FILE = __file__`.

kwkbtr: Or you can do the `library_path` resolution in the top level and do only the `include` in the function.

packages/qulacs/poetry.lock
```diff
@@ -482,4 +482,8 @@ files = [
 [metadata]
 lock-version = "2.0"
 python-versions = "^3.9.8"
+<<<<<<< HEAD
 content-hash = "b9c1eb307db47768bfad4037aeaae296209f68fd367965db8fb08f0d9573bbbc"
+=======
+content-hash = "805815cfb438e6c3401b20e61dd9a661abc9900b997a6bb8a836117a62c277dd"
+>>>>>>> main
```
kwkbtr: Conflict...

kwkbtr:
It seems that no docstrings are displayed on the documentation page.
Could you check it?
https://deploy-preview-33--quri-parts.netlify.app/quri_parts/itensor/quri_parts.itensor.html

speed1313:
> It seems that no docstrings are displayed on the documentation page. Could you check it? https://deploy-preview-33--quri-parts.netlify.app/quri_parts/itensor/quri_parts.itensor.html

While creating the doc, the doc creator needs the type information of funcitions. In this case, many functions in the ITensor package take  a juliacall type argument, such as juliacall.anyvalue as shown below.
```
def convert_circuit(
    circuit: NonParametricQuantumCircuit, qubit_sites: juliacall.VectorValue
) -> juliacall.VectorValue:
```
Hence, while doc ci is running, the juliacall dependency packages must be installed to know the type information of the juliacall.
To resolve this issue, I added the following statement to doc.yml:
```
    - run: poetry run julia -e 'using Pkg; Pkg.add("PythonCall")'
```
After making this change, the problem was resolved.

kwkbtr (approved):
👍

PR #37, opened 2023-03-20
Param shift grad estimator

tanan:
- add parameter shift gradient estimator to make VQE easier to use.
- remove `mapping_and_raw_circuit` function and add `param_mapping` and `primitive_circuit` function.
- move `Parameter` class from `circuit_parametric.py` to `parameter.py` to avoid a circular import.


packages/circuit/quri_parts/circuit/circuit_parametric.py
```diff
@@ -95,6 +75,34 @@ def get_mutable_copy(self) -> "MutableUnboundParametricQuantumCircuitProtocol":
...
+        \begin{align}
+            U_1(f(\theta_1, \theta_2)) U_2(g(\theta_1, \theta_2))
+        \end{align}
+
+        the primitive circuit should be as the following:
+
+        .. math::
+        \begin{align}
+            U_1(\psi_1) U_2(\psi_2)
+        \end{align}
```
kwkbtr: Math equations are not rendered correctly.
https://deploy-preview-37--quri-parts.netlify.app/quri_parts/circuit/quri_parts.circuit.circuit_parametric.html#quri_parts.circuit.circuit_parametric.UnboundParametricQuantumCircuitProtocol.primitive_circuit
Probably an extra indent should be added to each line inside `.. math::` directive.
```suggestion
        .. math::
            \begin{align}
                U_1(f(\theta_1, \theta_2)) U_2(g(\theta_1, \theta_2))
            \end{align}

        the primitive circuit should be as the following:

        .. math::
            \begin{align}
                U_1(\psi_1) U_2(\psi_2)
            \end{align}
```

packages/core/quri_parts/core/state/state_vector_parametric.py
```diff
@@ -53,6 +53,18 @@ def __repr__(self) -> str:
...
         return self._n_qubits
 
+    def with_primitive_circuit(self) -> "ParametricQuantumStateVector":
+        """Returns a new state with the primitive circuit applied. The
+        primitive circuit means that each gate in the circuit has an
+        independent parameter even if those in the original circuit depend on
+        the same parameters.
+
+        The original state is not changed.
+        """
```
kwkbtr: Probably this docstring needs to be updated in the same way as docstring of `ParametricCircuitQuantumState.with_primitive_circuit()`.

packages/circuit/quri_parts/circuit/circuit_parametric.py
tanan: fixed

packages/core/quri_parts/core/state/state_vector_parametric.py
tanan: fixed

kwkbtr (approved):
👍

PR #40, opened 2023-03-23
AllSinglesDoubles ansatz

toru4838:
Added AllSinglesDoubles ansatz, a parametric quantum circuit consists of single excitation and double excitation.

The implementation is based on PennyLane's AllSinglesDoubles. See the document https://docs.pennylane.ai/en/stable/code/api/pennylane.AllSinglesDoubles.html

### DIFF
--- a/packages/chem/quri_parts/chem/utils/excitations.py
+++ b/packages/chem/quri_parts/chem/utils/excitations.py
@@ -59,7 +59,7 @@ def add_single_excitation_circuit(
     param_fn: ParameterOrLinearFunction,
 ) -> LinearMappedUnboundParametricQuantumCircuit:
     r"""Add a particle-conserving single excitation circuit to the given
-    :attr:`circuit` implemented as a givens rotation :math:`G(\theta)`
+    :attr:`circuit` implemented as a Givens rotation :math:`G(\theta)`
 
     .. math::
         \begin{align}
--- a/packages/chem/quri_parts/chem/utils/excitations.py
+++ b/packages/chem/quri_parts/chem/utils/excitations.py
@@ -19,14 +19,14 @@ from quri_parts.circuit import (
     ParameterOrLinearFunction,
 )
 
-#: Represents the set of orbital indices involved in excitation. The first element is
-#: the index of an occupied orbital and the second element is the index of an
-#: unoccupied index.
+#: Alias of ``tuple[int, int]`` which represents the set of orbital indices involved in
+#: excitation. The first element is the index of an occupied orbital and the second
+#: element is the index of an unoccupied index.
 SingleExcitation: TypeAlias = tuple[int, int]
 
-#: Represents the set of orbital indices involved in excitation. The first and second
-#: element is the indices of occupied orbitals and the third and fourth element is the
-#: indices of unoccupied orbitals.
+#: Alias of ``tuple[int, int, int, int]`` which represents the set of orbital indices
+#: involved in excitation. The first and second element is the indices of occupied
+#: orbitals and the third and fourth element is the indices of unoccupied orbitals.
 DoubleExcitation: TypeAlias = tuple[int, int, int, int]
 
 
--- a/packages/chem/quri_parts/chem/utils/excitations.py
+++ b/packages/chem/quri_parts/chem/utils/excitations.py
@@ -77,7 +77,7 @@ def add_single_excitation_circuit(
         p_fn = {param: 0.5 * val for param, val in param_fn.items()}
 
     circuit.add_CNOT_gate(*excitation_indices)
-    _add_controlled_Y_gate(circuit, excitation_indices[1], excitation_indices[0], p_fn)
+    _add_controlled_RY_gate(circuit, excitation_indices[1], excitation_indices[0], p_fn)
     circuit.add_CNOT_gate(*excitation_indices)
     return circuit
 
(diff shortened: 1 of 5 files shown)


packages/chem/quri_parts/chem/utils/excitations.py
```diff
@@ -0,0 +1,164 @@
...
+
+#: Represents the set of orbital indices involved in excitation. The first element is
+#: the index of an occupied orbital and the second element is the index of an
+#: unoccupied index.
+SingleExcitation: TypeAlias = tuple[int, int]
+
+#: Represents the set of orbital indices involved in excitation. The first and second
+#: element is the indices of occupied orbitals and the third and fourth element is the
+#: indices of unoccupied orbitals.
+DoubleExcitation: TypeAlias = tuple[int, int, int, int]
```
kwkbtr: In current documentation it is not shown that these are type aliases of tuple, so it may be better to mention it explicitly.

https://deploy-preview-40--quri-parts.netlify.app/quri_parts/chem/quri_parts.chem.utils.excitations.html

packages/chem/quri_parts/chem/utils/excitations.py
```diff
@@ -0,0 +1,164 @@
...
+    return singles, doubles
+
+
+def add_single_excitation_circuit(
+    circuit: LinearMappedUnboundParametricQuantumCircuit,
+    excitation_indices: SingleExcitation,
+    param_fn: ParameterOrLinearFunction,
+) -> LinearMappedUnboundParametricQuantumCircuit:
+    r"""Add a particle-conserving single excitation circuit to the given
+    :attr:`circuit` implemented as a givens rotation :math:`G(\theta)`
```
kwkbtr:
```suggestion
    :attr:`circuit` implemented as a Givens rotation :math:`G(\theta)`
```

packages/chem/quri_parts/chem/utils/excitations.py
```diff
@@ -0,0 +1,164 @@
...
+    return circuit
+
+
+def add_double_excitation_circuit(
+    circuit: LinearMappedUnboundParametricQuantumCircuit,
+    excitation_indices: DoubleExcitation,
+    param_fn: ParameterOrLinearFunction,
+) -> LinearMappedUnboundParametricQuantumCircuit:
+    r"""Add a particle-conserving double excitation circuit to the given
+    :attr:`circuit` implemented as a extended givens rotation :math:`G^2(\theta)` which
```
kwkbtr:
```suggestion
    :attr:`circuit` implemented as a extended Givens rotation :math:`G^2(\theta)` which
```

packages/chem/quri_parts/chem/utils/excitations.py
```diff
@@ -0,0 +1,164 @@
...
+    doubles = []
+    for r in range(n_fermions):
+        for p in range(n_fermions, n_spin_orbitals):
+            if sz[p] - sz[r] == delta_sz:
+                singles.append((r, p))
+    for s in range(n_fermions - 1):
+        for r in range(s + 1, n_fermions):
+            for q in range(n_fermions, n_spin_orbitals - 1):
+                for p in range(q + 1, n_spin_orbitals):
+                    if sz[p] + sz[q] - sz[r] - sz[s] == delta_sz:
```
kwkbtr, line 50: I'm a bit worried about an error coming from numerical precision when comparing float values.
Should we use `int` for `delta_sz`?

packages/chem/quri_parts/chem/utils/excitations.py
```diff
@@ -0,0 +1,164 @@
...
+    circuit.add_CNOT_gate(excitation_indices[2], excitation_indices[0])
+    circuit.add_H_gate(excitation_indices[0])
+    circuit.add_H_gate(excitation_indices[3])
+    circuit.add_CNOT_gate(excitation_indices[0], excitation_indices[2])
+    circuit.add_CNOT_gate(excitation_indices[2], excitation_indices[3])
+
+    return circuit
+
+
+def _add_controlled_Y_gate(
```
kwkbtr: Is `_add_controlled_RY_gate` more appropriate name?

packages/chem/tests/chem/ansatz/test_all_singles_doubles.py
```diff
@@ -0,0 +1,63 @@
...
+from quri_parts.chem.ansatz.all_singles_doubles import AllSinglesDoubles
+from quri_parts.chem.utils.excitations import (
+    add_double_excitation_circuit,
+    add_single_excitation_circuit,
+    excitations,
+)
+from quri_parts.circuit import LinearMappedUnboundParametricQuantumCircuit
+
+
+def test_all_singles_doubles() -> None:
```
kwkbtr, line 20: This test case repeats the tested code.
I think it should be written more explicitly so that you can confirm correctness at a glance.
cf. https://github.com/QunaSys/quri-parts/blob/main/packages/algo/tests/algo/ansatz/test_hardware_efficient.py
(I don't think it's urgent, you can do it later in another pull request)

packages/chem/quri_parts/chem/utils/excitations.py
kwkbtr: But the values are always multiples of 0.5, so there should be no numerical errors.
So I think we can keep it as it is!

packages/chem/tests/chem/ansatz/test_all_singles_doubles.py
toru4838: Agree. I will create an issue so that I will not forget about it

kwkbtr (approved):
👍

PR #44, opened 2023-03-23
UCCSD ansatz

toru4838:
Added UCCSD, a commonly used ansatz for quantum chemistry calculation with VQE.
Note that current implementation only supports the excitations from HF state, i.e. base configuration is created from n_spin_orbitals and n_fermions and users cannot decide the occupied orbitals.

### DIFF
--- a/packages/openfermion/quri_parts/openfermion/ansatz/uccsd.py
+++ b/packages/openfermion/quri_parts/openfermion/ansatz/uccsd.py
@@ -1,0 +1,161 @@
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#      http://www.apache.org/licenses/LICENSE-2.0
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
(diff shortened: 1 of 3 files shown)


packages/openfermion/quri_parts/openfermion/ansatz/uccsd.py
```diff
@@ -0,0 +1,160 @@
...
+    only supports singlet state and the occupied orbitals are the lowest
+    :attr:`n_fermions` spin orbitals.
+
+    Args:
+        n_spin_orbitals: Number of spin orbitals.
+        n_fermions: Number of fermions.
+        fermion_qubit_mapping: Mapping from :class:`FermionOperator` to
+          :class:`Operator`
+        trotter_number: Number for first-order Trotter product formula.
+        use_single: If ``True``, single-excitation gates are applied.
```
kwkbtr:
```suggestion
        use_singles: If ``True``, single-excitation gates are applied.
```

packages/openfermion/tests/openfermion/ansatz/test_uccsd.py
kwkbtr, line 1: It seems that there are many test cases in each test function.
It will be more readable if each test function, with a good test title, contains only one test case.

kwkbtr (approved):
LGTM, except:
- A minor typo
- Test case reorganization
- More detailed docstring describing assumptions made in the implementation

I think all of them can be addressed later.

kwkbtr (approved):

PR #48, opened 2023-03-29
Fix location of NOTICE file in wheel packages

kwkbtr:
Workaround for https://github.com/python-poetry/poetry/issues/7153
Since `NOTICE` file from multiple sub packages are placed in the same location (just under `site-packages`), uninstallation of the packages fails (`pip` complains that it cannot find `NOTICE` file, which was removed by uninstallation of other packages).

### DIFF
--- a/.github/workflows/package.yml
+++ b/.github/workflows/package.yml
@@ -32,12 +32,14 @@ jobs:
     - run: |
         mkdir -p dist
         poetry build
-        for pkg in packages/*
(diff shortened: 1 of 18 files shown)


kwkbtr:
I checked the generated wheel files and confirmed the `NOTICE` file location:

- For the root package: `quri_parts/NOTICE`
- For each sub package: `quri_parts/{pkg}/NOTICE`

PR #50, opened 2023-03-29
SqrtX gate support in quri-parts-openqasm

snuffkin:
SqrtX(sx) is available in both QURI Parts and OpenQASM 3.0, so quri-parts-openqasm supports it.

stdgates.inc
https://github.com/openqasm/openqasm/blob/main/examples/stdgates.inc#L26


kwkbtr:
Could you resolve the conflict?

PR #52, opened 2023-03-30
Transpiler for OpenQASM decomposing multi-Pauli (rotation) gates

kwkbtr:
Since multi-Pauli (rotation) gates are not defined in [stdgates.inc](https://github.com/openqasm/openqasm/blob/4ad00941a8d863fd462d2c1757e282b8429bfe32/examples/stdgates.inc), it should be decomposed before converted to OpenQASM.

### DIFF
--- a/packages/openqasm/quri_parts/openqasm/circuit/__init__.py
+++ b/packages/openqasm/quri_parts/openqasm/circuit/__init__.py
@@ -9,7 +9,7 @@
 # limitations under the License.
 
 import io
-from typing import TYPE_CHECKING, Mapping
+from typing import TYPE_CHECKING, Callable, Mapping
 
 from quri_parts.circuit import gate_names
 from quri_parts.circuit.gate_names import (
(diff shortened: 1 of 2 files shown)


packages/openqasm/quri_parts/openqasm/circuit/__init__.py
```diff
@@ -30,6 +35,19 @@
...
+class OpenQASMTranspiler(SequentialTranspiler):
+    """CircuitTranspiler which transpiles gates not supported by OpenQASM to
+    supported ones."""
+
+    def __init__(self) -> None:
+        transpilers = [
+            PauliDecomposeTranspiler(),
+            PauliRotationDecomposeTranspiler(),
+        ]
+        super().__init__(transpilers)
```
lqtmirage: If there is no special processing, I think it would be acceptable to create a SequentialTranspiler directly.
```python
OpenQASMTranspiler: Callable[[], CircuitTranspiler] = lambda: SequentialTranspiler(
    [PauliDecomposeTranspiler(), PauliRotationDecomposeTranspiler()]
)
```

lqtmirage (approved):

PR #60, opened 2023-04-14
Update qulacs

kwkbtr:


dchung0741:
I tried to fix the typing problems as much as I can without casting and ignoring.
Here are some comments:

1. For lines related to the `DenseMatrix` class and those using `.get_matrix()`, `.get_vector()` methods, they are related to the errors in the `.ipy` files in `qulacs`, so I still had to resort to using `# type: ignore`
2. Functions like `CPTP` and `Probabilistic` has no return type hint, so I had to use `cast`.
3. For the gate mapping dictionaries inside `quri_parts.qulacs.test_convert_circuit`. I had to add some `TypeAlias` for the gate factory classes inside `quri_parts.circuit.gates.py`. I'm not sure if it is the proper file to place the definitions in.
4. When the `.load()` method for qulacs states is called, the argument needs to be converted to a list.

kwkbtr:
- Let's use `cast(list[int], ...)` instead of `list()` to avoid actual object conversion. It may be good to define a function that only does the cast so that we can remove it easily when type annotation of qulacs gets fixed.
- Please add a comment to the unavoidable `# type: ignore` describing the reason.
- It may be better to use `_dense_matrix_gate_qulacs()` in `circuit/__init__.py` instead of `DenseMatrix()`.

> 3. For the gate mapping dictionaries inside `quri_parts.qulacs.test_convert_circuit`. I had to add some `TypeAlias` for the gate factory classes inside `quri_parts.circuit.gates.py`. I'm not sure if it is the proper file to place the definitions in.

I don't understand the reason of this. Can you elaborate on that?

dchung0741:
> * Let's use `cast(list[int], ...)` instead of `list()` to avoid actual object conversion. It may be good to define a function that only does the cast so that we can remove it easily when type annotation of qulacs gets fixed.

I added 
```python
Numerics = TypeVar('Numerics', int, float, complex)
def cast_to_list(int_sequence: Sequence[Numerics]) -> list[Numerics]:
    return cast(list[Numerics], int_sequence)
```
inside `quri_parts.qulacs.__init__.py` to do the job. (Not pushed yet.)

> * Please add a comment to the unavoidable `# type: ignore` describing the reason.

Do you mean comment inside the .py files?

> * It may be better to use `_dense_matrix_gate_qulacs()` in `circuit/__init__.py` instead of `DenseMatrix()`.

Sorry do you mean `_dense_matrix_gate_qulacs()` instead of # type: ignore or casting?
I'm unsure because `DenseMatrix()` and `_dense_matrix_gate_qulacs()` takes in different number of arguments.

> 
> > 3. For the gate mapping dictionaries inside `quri_parts.qulacs.test_convert_circuit`. I had to add some `TypeAlias` for the gate factory classes inside `quri_parts.circuit.gates.py`. I'm not sure if it is the proper file to place the definitions in.
> 
> I don't understand the reason of this. Can you elaborate on that?

Sorry, I just realized that the original error was with the second argument of `Mapping`. I have removed the unnecessary TypeAlias definitions. (Not pushed yet)

kwkbtr:
> I added
> 
> ```python
> Numerics = TypeVar('Numerics', int, float, complex)
> def cast_to_list(int_sequence: Sequence[Numerics]) -> list[Numerics]:
>     return cast(list[Numerics], int_sequence)
> ```
> 
> inside `quri_parts.qulacs.__init__.py` to do the job. (Not pushed yet.)

👍 

> > * Please add a comment to the unavoidable `# type: ignore` describing the reason.
> 
> Do you mean comment inside the .py files?

Yes, so that we can see the reason later and can remove them when the upstream (qulacs) issue is fixed.

> > * It may be better to use `_dense_matrix_gate_qulacs()` in `circuit/__init__.py` instead of `DenseMatrix()`.
> 
> Sorry do you mean `_dense_matrix_gate_qulacs()` instead of # type: ignore or casting? I'm unsure because `DenseMatrix()` and `_dense_matrix_gate_qulacs()` takes in different number of arguments.

I mean `qulacs.gate.DenseMatrix(...)` creates a `DenseMatrix` object, and `_dense_matrix_gate_qulacs()` does the exactly the same with the casting and `# type: ignore`, so it should be possible to replace `DenseMatrix()` with it. You can modify its signature if necessary.

> > > 3. For the gate mapping dictionaries inside `quri_parts.qulacs.test_convert_circuit`. I had to add some `TypeAlias` for the gate factory classes inside `quri_parts.circuit.gates.py`. I'm not sure if it is the proper file to place the definitions in.
> > 
> > 
> > I don't understand the reason of this. Can you elaborate on that?
> 
> Sorry, I just realized that the original error was with the second argument of `Mapping`. I have removed the unnecessary TypeAlias definitions. (Not pushed yet)

👍

dchung0741:
Some summaries on the qulacs types that are causing the  problems

1. Second argument of `DenseMatrix` and `add_dense_matrix_gate` takes in `NDArray[complex, _Shape]` while `_Shape` should be the first argument.
2. `qulacs.QuantumState.load()` method and `qulacs.gate.Pauli()` takes in List[Numerics], which is too strict (Sequence[Numerics] should be enough).
3. `qulacs.gate.CPTP` and `qulacs.gate.Probabilistic` are missing return type annotations.

packages/qulacs/quri_parts/qulacs/circuit/__init__.py
```diff
@@ -92,6 +121,19 @@
 }
 
 
+def _dense_matrix_gate_qulacs(
+    t: Union[int, Sequence[int]], unitary_matrix: ArrayLike
+) -> qulacs.QuantumGateBase:
+    # This cast is workaround for too strict type annotation of Qulacs
+    # t = cast(list[int], gate.target_indices)
```
kwkbtr: OK, let's remove these commented lines.

packages/ionq/tests/ionq/circuit/transpile/test_ionq_native_transpiler.py
```diff
@@ -97,12 +97,18 @@ def ionq_circuit_state(
     qs_circuit = qulacs.QuantumCircuit(circuit.qubit_count)
     for gate in circuit.gates:
         if gate.name in ionq_gates_1:
+            # the second argument of `.add_dense_matrix_gate()`
+            # takes in NDArray[complex128, _Shape]
+            # (_Shape should be the first argument)
```
kwkbtr: I think the comment in `qulacs/circuit/__init__.py` is clearer:

```suggestion
            # We need to disable type check due to an error in qulacs type annotation
            # https://github.com/qulacs/qulacs/issues/537
```

packages/ionq/tests/ionq/circuit/transpile/test_ionq_native_transpiler.py
```diff
@@ -97,12 +97,18 @@ def ionq_circuit_state(
...
             qs_circuit.add_dense_matrix_gate(
-                gate.target_indices[0], ionq_gate_matrix(gate)
+                gate.target_indices[0], ionq_gate_matrix(gate)  # type: ignore
             )
         elif gate.name in ionq_gates_2:
+            # the second argument of `.add_dense_matrix_gate()`
+            # takes in NDArray[complex128, _Shape]
+            # (_Shape should be the first argument)
```
kwkbtr: I think the comment in `qulacs/circuit/__init__.py` is clearer:

```suggestion
            # We need to disable type check due to an error in qulacs type annotation
            # https://github.com/qulacs/qulacs/issues/537
```

packages/qulacs/quri_parts/qulacs/circuit/noise/gate_converter.py
```diff
@@ -9,36 +9,57 @@
...
 
 
 def create_pauli_noise_gate(
     qubits: Sequence[int], noise: PauliNoise
 ) -> QuantumGateBase:
     qubits = noise.qubit_indices
     pauli_list, prob_list = noise.pauli_list, list(noise.prob_list)
+    # have to convert to list(noise.prob_list), otherwise mypy throws:
+    # Argument 1 to "append" of "list" has incompatible type "float"; expected "int"
+    # Even if the `reveal_type(cast_to_list(prob_list))` shows it is a list of floats.
```
kwkbtr: I don't understand what this comment refers to.
If it is about `list(noise.prob_list)` above, it is because `noise.prob_list` is a tuple and need to be converted into a list.

packages/qulacs/quri_parts/qulacs/simulator.py
```diff
@@ -44,11 +46,13 @@ def run_circuit(
...
     qulacs_state = ql.QuantumState(circuit.qubit_count)
-    qulacs_state.load(init_state)
+    qulacs_state.load(cast_to_list(init_state))
 
     qulacs_cicuit = convert_circuit(circuit)
     qulacs_cicuit.update_quantum_state(qulacs_state)
 
-    new_state_vector: NDArray[cfloat] = qulacs_state.get_vector()
+    # The `.get_matrix()` method returns NDArray[complex128, _Shape]
+    # (_Shape should be the first argument)
```
kwkbtr: I think this is clearer:
```suggestion
    # We need to disable type check due to an error in qulacs type annotation
    # https://github.com/qulacs/qulacs/issues/537
```

packages/qulacs/tests/qulacs/circuit/test_noise_model_qulacs.py
```diff
@@ -44,8 +44,10 @@ def gate_info(
             g.get_control_index_list(),
         )
 
+    # the `.get_matrix()` method returns NDArray[complex128, _Shape]
+    # (_Shape should be the first argument)
```
kwkbtr: I think this is clearer:
```suggestion
    # We need to disable type check due to an error in qulacs type annotation
    # https://github.com/qulacs/qulacs/issues/537
```

packages/qulacs/tests/qulacs/circuit/test_noise_model_qulacs.py
```diff
@@ -247,4 +249,6 @@ def test_convert_kraus_cptp() -> None:
         kdg = k.conj().T
         cptp += k @ rho @ kdg  # K_i rho K_i^dagger
 
-    assert np.allclose(cptp, matrix.get_matrix())
+    # the `.get_matrix()` method returns NDArray[complex128, _Shape]
+    # (_Shape should be the first argument)
```
kwkbtr: I think this is clearer:
```suggestion
    # We need to disable type check due to an error in qulacs type annotation
    # https://github.com/qulacs/qulacs/issues/537
```

packages/qulacs/tests/qulacs/circuit/test_convert_circuit.py
```diff
@@ -121,23 +124,30 @@ def test_convert_rotation_gate() -> None:
     # Check rotation angle sign
     c = np.cos(np.pi / 4)
     s = np.sin(np.pi / 4)
+
+    # the `.get_matrix()` method returns NDArray[complex128, _Shape]
+    # (_Shape should be the first argument)
```
kwkbtr: I think this is clearer:
```suggestion
    # We need to disable type check due to an error in qulacs type annotation
    # https://github.com/qulacs/qulacs/issues/537
```

And it may be sufficient to comment only once in one function.

packages/qulacs/tests/qulacs/circuit/test_convert_circuit.py
```diff
@@ -155,21 +165,25 @@ def test_convert_u_gate() -> None:
 def test_convert_pauli_gate() -> None:
     g = gates.Pauli((11, 7, 13), (2, 3, 1))
     converted = convert_gate(g)
-    expected = qulacs.gate.Pauli((11, 7, 13), (2, 3, 1))
+    expected = qulacs.gate.Pauli(cast_to_list((11, 7, 13)), cast_to_list((2, 3, 1)))
```
kwkbtr: In this case we can simply use lists.
(My past comment about favoring casting is to reduce performance overhead of object conversion when a tuple is already given)

packages/qulacs/tests/qulacs/circuit/test_convert_circuit.py
```diff
@@ -155,21 +165,25 @@ def test_convert_u_gate() -> None:
...
     assert gates_equal(converted, expected)
 
 
 def test_convert_pauli_rotation_gate() -> None:
     g = gates.PauliRotation((11, 7, 13), (2, 3, 1), 0.125)
     converted = convert_gate(g)
-    expected = qulacs.gate.PauliRotation((11, 7, 13), (2, 3, 1), -0.125)
+    expected = qulacs.gate.PauliRotation(
+        cast_to_list((11, 7, 13)), cast_to_list((2, 3, 1)), -0.125
```
kwkbtr: In this case we can simply use lists.
(My past comment about favoring casting is to reduce performance overhead of object conversion when a tuple is already given)

packages/qulacs/tests/qulacs/circuit/test_convert_circuit.py
```diff
@@ -155,21 +165,25 @@ def test_convert_u_gate() -> None:
...
+    )
     assert gates_equal(converted, expected)
 
     # Check rotation angle sign
     c = np.cos(np.pi / 4)
     s = np.sin(np.pi / 4)
+    # the `.get_matrix()` method returns NDArray[complex128, _Shape]
+    # (_Shape should be the first argument)
```
kwkbtr: I think this is clearer:
```suggestion
    # We need to disable type check due to an error in qulacs type annotation
    # https://github.com/qulacs/qulacs/issues/537
```

packages/qulacs/tests/qulacs/circuit/test_convert_circuit.py
```diff
@@ -222,7 +236,9 @@ def test_convert_parametric_circuit() -> None:
         qulacs.gate.CNOT(0, 2),
         qulacs.gate.ParametricRZ(2, 0.0),
         qulacs.gate.RX(0, -0.125),
-        qulacs.gate.ParametricPauliRotation((0, 1, 2), (1, 2, 3), 0.0),
+        qulacs.gate.ParametricPauliRotation(
+            cast_to_list((0, 1, 2)), cast_to_list((1, 2, 3)), 0.0
```
kwkbtr: In this case we can simply use lists.
(My past comment about favoring casting is to reduce performance overhead of object conversion when a tuple is already given)

packages/qulacs/tests/qulacs/circuit/test_convert_circuit.py
```diff
@@ -253,7 +269,9 @@ def test_convert_linear_mapped_parametric_circuit() -> None:
...
```
kwkbtr: In this case we can simply use lists.
(My past comment about favoring casting is to reduce performance overhead of object conversion when a tuple is already given)

packages/qulacs/quri_parts/qulacs/circuit/__init__.py
```diff
@@ -117,12 +167,20 @@ def convert_gate(
             *gate.control_indices, *gate.target_indices
         )
     elif is_multi_qubit_gate_name(gate.name):
-        neg_params = (-p for p in gate.params)
-        return _multi_pauli_gate_qulacs[gate.name](
-            gate.target_indices, gate.pauli_ids, *neg_params
-        )
+        # These cast are workaround for too strict type annotation of Qulacs
+        target_indices = cast(list[int], gate.target_indices)
+        pauli_ids = cast(list[int], gate.pauli_ids)
```
kwkbtr: Let's use `cast_to_list()`.

packages/qulacs/quri_parts/qulacs/circuit/__init__.py
```diff
@@ -172,8 +230,11 @@ def param_mapper(s: Sequence[float]) -> Sequence[float]:
             elif gate.name == gate_names.ParametricRZ:
                 qulacs_circuit.add_parametric_RZ_gate(gate.target_indices[0], 0)
             elif gate.name == gate_names.ParametricPauliRotation:
+                # These cast are workaround for too strict type annotation of Qulacs
+                target_indices = cast(list[int], gate.target_indices)
```
kwkbtr: Let's use `cast_to_list()`.

packages/qulacs/quri_parts/qulacs/__init__.py
```diff
@@ -18,3 +19,9 @@
 QulacsParametricStateT: TypeAlias = Union[
     ParametricCircuitQuantumState, ParametricQuantumStateVector
 ]
+
+Numerics = TypeVar("Numerics", int, float, complex)
+
+
+def cast_to_list(int_sequence: Union[Sequence[Numerics], ArrayLike]) -> list[Numerics]:
+    return cast(list[Numerics], int_sequence)
```
kwkbtr: Let's add a docstring to describe the purpose of this function.

packages/qulacs/quri_parts/qulacs/circuit/noise/gate_converter.py
dchung0741: This is explaining why I need to do use `list(noise.prob_list)` on that line instead of doing
`cast_to_list(prob_list).append(prob_identity)` a couple of lines later.

If I do `cast_to_list(prob_list).append(prob_identity)`, `mypy` thinks `cast_to_list(prob_list)` is a list of integers while it is obviously a list of floats, thus throwing the error message `Argument 1 to "append" of "list" has incompatible type "float"; expected "int"`.

I think I'm going to delete the comments and leave the line `list(noise.prob_list)`

kwkbtr: `list(noise.prob_list)` existed before this PR by a different reason (which I described in the comment).
I'm not sure about the mypy error, but doing `cast_to_list(prob_list).append(...)` instead of `list(prob_list)` will throw in runtime since `prob_list` can be a tuple.

dchung0741: I see. I think I confused myself when removing the redundant `list` calls I added.
I have reverted it back.

packages/qulacs/quri_parts/qulacs/__init__.py
```diff
@@ -18,3 +19,11 @@
 QulacsParametricStateT: TypeAlias = Union[
     ParametricCircuitQuantumState, ParametricQuantumStateVector
 ]
+
+Numerics = TypeVar("Numerics", int, float, complex)
+
+
+def cast_to_list(int_sequence: Union[Sequence[Numerics], ArrayLike]) -> list[Numerics]:
+    """Cast a sequence of numerics (int, float, complex) or an array to a list
+    of the same type."""
```
kwkbtr: Could you describe the purpose of this function (workaround for too strict type annotation of Qulacs)?

kwkbtr (commented):
LGTM, except:
- 1 new comment (improvement of docstring)
- 2 comments from the last review (replacing cast() with cast_to_list())

packages/qulacs/quri_parts/qulacs/circuit/__init__.py
dchung0741: Modified

packages/qulacs/quri_parts/qulacs/circuit/__init__.py
dchung0741: Modified

packages/qulacs/quri_parts/qulacs/__init__.py
dchung0741: The current docstring is:
```python
"""Cast a sequence of numerics (int, float, complex) or an array to a list
of the same type.

This is a workaround for too strict type annotation of Qulacs
"""
```

kwkbtr (commented):
👍

PR #83, opened 2023-05-08
Add create_concurrent_parametric_estimator

r-imai-67:
This PR is to add a helper method that creates `ConcurrentParametricQuantumEstimator` from `ParametricQuantumEstimator`.

### DIFF
--- a/packages/core/quri_parts/core/estimator/__init__.py
+++ b/packages/core/quri_parts/core/estimator/__init__.py
@@ -185,6 +185,23 @@ def create_parametric_estimator(
     return parametric_estimator
 
 
+def create_concurrent_parametric_estimator(
+    parametric_estimator: ParametricQuantumEstimator[_ParametricStateT],
+) -> ConcurrentParametricQuantumEstimator[_ParametricStateT]:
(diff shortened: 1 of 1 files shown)


packages/core/quri_parts/core/estimator/__init__.py
```diff
@@ -185,6 +185,24 @@ def parametric_estimator(
...
+
+    def concurrent_parametric_estimator(
+        operator: Estimatable,
+        state: _ParametricStateT,
+        seq_of_params: Sequence[Sequence[float]],
+    ) -> list[Estimate[complex]]:
+        ret = []
+        for params in seq_of_params:
+            ret += [parametric_estimator(operator, state, params)]
+        return ret
```
kwkbtr: I'm a bit worried about efficiency of list concatenation.
Using list comprehension might be better?

r-imai-67: Thank you for your comment!
I've replaced that part with a list comprehension.

kwkbtr (approved):

PR #85, opened 2023-05-10
Fill diagonal elements of error_matrix from gradient estimators

r-imai-67:
In the current implementation, gradient estimators do not return error_matrix.
This PR contains changes that make estimators return a matrix where the diagonal elements are filled with its error.
Note that non-diagonal elements are filled with zeros without calculation.

### DIFF
--- a/packages/core/quri_parts/core/estimator/gradient.py
+++ b/packages/core/quri_parts/core/estimator/gradient.py
@@ -12,6 +12,8 @@ from collections.abc import Sequence
 from dataclasses import dataclass
 from typing import Optional, TypeVar, Union, cast
 
+import numpy as np
+
 from quri_parts.circuit.parameter_mapping import LinearParameterMapping
 from quri_parts.circuit.parameter_shift import ShiftedParameters
 from quri_parts.core.estimator import (
@@ -61,8 +63,7 @@ def numerical_gradient_estimates(
 
     Returns:
         The estimated values (can be accessed with :attr:`.values`) with errors
-        of estimation (can be accessed with :attr:`.error_matrix`). Currently,
-        :attr:`.error_matrix` returns `None`.
+        of estimation (can be accessed with :attr:`.error_matrix`).
     """
 
     v = []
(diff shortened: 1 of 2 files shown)


packages/core/quri_parts/core/estimator/gradient.py
```diff
@@ -77,11 +79,15 @@ def numerical_gradient_estimates(
...
+    err_diag = []
     for i in range(len(params)):
         d = estimates[2 * i].value - estimates[2 * i + 1].value
         grad.append(d / delta)
-
-    return _Estimates(grad, None)
+        var = (
+            (estimates[2 * i].error ** 2) + (estimates[2 * i + 1].error ** 2)
+        ) / delta
+        err_diag.append(np.sqrt(var))
```
kwkbtr: This seems to be $\sqrt{(e_1^2+e_2^2)/\delta}$, but isn't $\sqrt{(e_1^2+e_2^2)}/\delta$ correct?

packages/core/quri_parts/core/estimator/gradient.py
```diff
@@ -62,7 +64,7 @@ def numerical_gradient_estimates(
     Returns:
         The estimated values (can be accessed with :attr:`.values`) with errors
         of estimation (can be accessed with :attr:`.error_matrix`). Currently,
-        :attr:`.error_matrix` returns `None`.
+        non-diagonal elements of :attr:`.error_matrix` are filled with zeros.
```
kwkbtr: Saying "are filled with zeros" sounds to me like "we don't know the exact values but we put zeros at the moment", but I guess the covariances are exactly zero in this case since each element is estimated indepedently.

packages/core/quri_parts/core/estimator/gradient.py
```diff
@@ -171,13 +177,17 @@ def parameter_shift_gradient_estimates(
     estimates_dict = dict(zip(gate_params_list, estimates))
 
     grad = []
+    err_diag = []
     for params_and_coefs in shifted_params_and_coefs:
         g = 0.0 + 0.0j
+        var = 0.0
         for p, c in params_and_coefs:
             g += estimates_dict[p].value * c
+            var += (estimates_dict[p].error ** 2) * abs(c)
```
kwkbtr: I guess $|c|$ should also be squared:
```suggestion
            var += (estimates_dict[p].error ** 2) * abs(c) ** 2
```

packages/core/quri_parts/core/estimator/gradient.py
r-imai-67: Thank you for pointing this out!
I misremembered the propagation rule of error.
The latter one is correct.

kwkbtr (approved):

packages/core/quri_parts/core/estimator/gradient.py
r-imai-67: I was a bit confused but, as you guessed, the covariance are exactly zero.
So I've removed the sentence.

r-imai-67:
@kwkbtr 
Thank you for reviewing!

PR #94, opened 2023-05-23
Nr chem general interface

dchung0741:
An interface for non-relativistic quantum chemistry which is able to generate full space and active space electron integrals.


dchung0741:
Let me fix the author issue. It seems like there are some problem with the setting when I switched my computer

dchung0741:
I tried to figure out how to correct the emails in the old commit but it seems a bit complicated because those commits are buried too deep.
I found this command which seems to rewrite the emails:

```git
git filter-branch --commit-filter '
    if [ "$GIT_COMMITTER_NAME" = "Daniel" ];
    then
        GIT_COMMITTER_EMAIL="chung@qunasys.com](mailto:chung@qunasys.com)";
        GIT_AUTHOR_EMAIL="[chung@qunasys.com](mailto:chung@qunasys.com)";
        git commit-tree "$@";
    else
        git commit-tree "$@";
    fi' HEAD
```
but I'm not confident because it seems to rewrite other peoples commits.
Is it good if I create a new branch with only 1 commit ahead of main that contains all the changes in this PR?

kwkbtr:
> but I'm not confident because it seems to rewrite other peoples commits.

I'm not sure if that command works or not either.

> Is it good if I create a new branch with only 1 commit ahead of main that contains all the changes in this PR?

Yes, it's fine.

PR #96, opened 2023-05-24
non-relativistic quantum chemistry

dchung0741:
An interface for non-relativistic quantum chemistry which is able to generate full space and active space electron integrals.


kwkbtr (approved):
Great work, thanks! 🎉

dchung0741:
Thanks!

PR #117, opened 2023-06-14
Fix UnitaryMatrix KAK decomposition and conversions to backends

lqtmirage:
- Corrected the relationship between the sequence of qubit indices and the unitary matrix when converting a `UnitaryMatrix` gate to Braket or Cirq.
- For each backend, added a test to compare the matrix representation of the converted `UnitaryMatrix` gate with that of the KAK-decomposed circuit.
- Bug fix in `TwoQubitUnitaryMatrixKAKTranspiler`. (Fixed a bug in which the qubit number contained a constant value, and also changed the specification for the assignment of qubits.)


kwkbtr (approved):
👍

PR #118, opened 2023-06-14
Add fusing rotation gates to the preset gate set decomposition paths

lqtmirage:
- Since the decomposition paths to prefix gate sets only transforms individual gates, it often yields a series of rotation gates. As a minimal optimization, added a path that fuses these rotation gates.
- No further optimizations were included in the decomposition paths so that the optimization can be controlled by the user.


kwkbtr (commented):
👍

kwkbtr (approved):

PR #122, opened 2023-06-17
Unify python & poetry install to composite action

y-yu:
- https://docs.github.com/en/actions/creating-actions/creating-a-composite-action

### DIFF
--- a/.github/actions/python-poetry-install/action.yml
+++ b/.github/actions/python-poetry-install/action.yml
@@ -23,4 +23,3 @@ runs:
     - uses: actions/setup-python@v4
       with:
         python-version: ${{ inputs.python-version }}
-        cache: 'poetry'
--- a/.github/actions/python-poetry-install/action.yml
+++ b/.github/actions/python-poetry-install/action.yml
@@ -1,0 +1,26 @@
+name: "Python & Poetry install and cache"
+description: "Python & Poetry install and cache"
+
+inputs:
+  poetry-version:
+    description: "Poetry version"
+    required: true
+  python-version:
+    description: "Python version"
+    required: true
+
+runs:
+  using: "composite"
+  steps:
+    - name: Install Poetry
+      env:
+        POETRY_VERSION: ${{ inputs.poetry-version }}
+      shell: bash
+      run: |
+        curl -sSL https://install.python-poetry.org | python -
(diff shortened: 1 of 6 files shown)


y-yu:

.github/actions/python-poetry-install/action.yml
```diff
@@ -0,0 +1,26 @@
...
+        POETRY_VERSION: ${{ inputs.poetry-version }}
+      shell: bash
+      run: |
+        curl -sSL https://install.python-poetry.org | python -
+        echo "$HOME/.poetry/bin" >> $GITHUB_PATH
+
+    - uses: actions/setup-python@v4
+      with:
+        python-version: ${{ inputs.python-version }}
+        cache: 'poetry'
```
y-yu: I enable the `cache: 'poetry'` option following https://github.com/actions/setup-python#caching-packages-dependencies 👀

.github/workflows/package.yml
```diff
@@ -12,16 +12,9 @@ jobs:
...
-      run: |
-        curl -sSL https://install.python-poetry.org | python -
-        echo "$HOME/.poetry/bin" >> $GITHUB_PATH
-        poetry self add "poetry-dynamic-versioning[plugin]"
-
-    - uses: actions/setup-python@v4
+    - uses: ./.github/actions/python-poetry-install
       with:
+        poetry-version: '1.4.0'
         python-version: '3.9'
```
kwkbtr: This needs to be moved after `actions/checkout`.

.github/actions/python-poetry-install/action.yml
kwkbtr: Note the following comment on https://github.com/QunaSys/quri-parts/pull/116:

> - In each workflow (`lint`, `typecheck`, `test` and `doc`), only necessary dependencies are installed with `poetry install --only [groups]`
>   - Since installed dependecies change depending on the workflow, I disabled caching of poetry virtualenv. (It may be safe to keep it enabled, but it is not likely that the caching provides much improvement on the execution time)

y-yu:
OK, it should be better not to cache for poetry so I remove the setting for `cache` of `action/setup-python`.

kwkbtr (approved):

kwkbtr:
Thank you!

PR #132, opened 2023-06-21
Yozu julia cache

yasuo-ozu:


kwkbtr (commented):
I think it would be better to use in the cache key a unique value like a hash of a project file, since the same cache will be used forever otherwise.
(It would not be a big problem, though, given that our Julia dependencies don't change much)

kwkbtr (approved):

PR #154, opened 2023-07-04
Inverse state mapper

kwkbtr:


packages/chem/quri_parts/chem/transforms/__init__.py
```diff
@@ -85,6 +91,28 @@ def get_state_mapper(
...
+
+        Args:
+            n_spin_orbitals:
+                The number of spin orbitals.
+            n_fermions:
+                The number of fermions considered when the qubit state is mapped.
+                Some mappings require this argument (e.g. symmetry-conserving
+                Bravyi-Kitaev transformation) while the others ignore it.
+            n_up_spins:
+                The number of spin orbitals.
```
kwkbtr: This docstring needs to be updated.

kwkbtr (commented):
Moving post-selection filter functions from `quri_parts.chem` to `quri_parts.openfermion` is a breaking change, so let's note it explicitly in pull request description.

dchung0741:
I haven't read the post selection functions yet, but I have one question about the updated interface for incorporating the inverse mappers.
It looks like the inverse mapper could take in the `n_up_spins` argument.
For the SCBK Hamiltonian mapping, it also needs to take in `n_up_spins` in order to compute the correct parity factor for the Hamiltonians.
So my question is: Should we add the `n_up_spins` argument to the `get_state_mapper`, `get_of_operator_mapper` methods as well?

kwkbtr:
> Should we add the `n_up_spins` argument to the `get_state_mapper`, `get_of_operator_mapper` methods as well?

I guess

- For `get_of_operator_mapper`: Yes
- For `get_state_mapper`: I think the number of up spins can be determined from the input (occupation numbers). But it is also the case for `n_fermions`, so I think it's good to add it for consistency.

packages/openfermion/quri_parts/openfermion/transforms/__init__.py
```diff
@@ -293,5 +340,28 @@ def get_state_mapper(
...
+
+    def augment_dropped_bits(
+        self,
+        bit_array: list[int],
+        n_spin_orbitals: int,
+        n_fermions: Optional[int] = None,
+    ) -> list[int]:
+        # Add two qubits dropped by the fermion-to-qubit mapping.
+        return bit_array + [0, 0]
+
```
dchung0741: I'm wondering if this method should be public?
My concern is that for SCBK inverse mapping, we do not rearrange the dropped qubit back to the M/2-th and M-th qubit with this method (M is the number of spin orbitals.) Would this cause user's confusions if they decide to use it as a public method?

dchung0741 (commented):
The code looks good to me overall. Just one slight concern.
Also, about the `n_up_spins` should I add it in my PR later?

packages/openfermion/quri_parts/openfermion/transforms/__init__.py
kwkbtr: Yeah, maybe we can make it private (or you can call it "protected", a term used in other programming languages like C++ or Java).

toru4838: I will make it protected in this PR

toru4838:
Thank you for your feedback!

> Also, about the n_up_spins should I add it in my PR later?

This PR only considers restricted (OF's) transformations, so we don't need `n_up_spins` for `get_operator_mapper()` and `get_state_mapper()` here. Let's add it in your PR later.

toru4838 (approved):

PR #166, opened 2023-07-12
Ignore patch version updates of botocore

kwkbtr:
Since botocore and boto3 are updated almost every day with patch version change, I want to ignore those updates to reduce number of dependabot pull requests.

cf. https://github.blog/changelog/2021-05-21-dependabot-version-updates-can-now-ignore-major-minor-patch-releases/


dchung0741 (approved):
👍

PR #168, opened 2023-07-12
Measurement support for `QuantumCircuit`

Zshan0:
Measurement support in the `QuantumCircuit` includes the following features:

### Addition of classical bits

`QuantumCircuitProtocol` has been extended with an addition of `cbit_count` that indicates the number of classical bits in the circuit. The interface for this is the same as `qubit_count`. All classes that follow the protocol now support classical bits as well. Classical bits can currently be used as the target for measurement results of qubits. They will later be useful for classical conditional operators.

### Measurement operator

As an exception to `QuantumGate` being unitary, `Measurement` has been added as a `QuantumGate` that is not unitary. This gate can be used as any other `QuantumGate` except it requires a sequence of classical indices along with target indices to indicate the final address of measurement results. The gate can be constructed using `MutableQuantumCircuitProtocol.measure` that supports the interface of single or multiple qubit measurement.

Note that `Measurement` is a multi-qubit gate that can operate on any number of qubits.

### `cbit_count` addition to `QuantumCircuit`

The addition of `cbit_count` as a parameter for initialization of `QuantumCircuit` will be a cause of a breaking change in the next version of `quri-parts` as `gates` parameter has been pushed from 2nd to 3rd position. Currently, we accept `QuantumCircuit(qubit_count, gates)` at runtime, but this comes with a deprecation warning.

### Qiskit Transpilation

To ensure that the new operation is easy to transpile, the qiskit transpilation has been implemented and is fully supported. The support for measurement ops could be added other SDKs as well.


kwkbtr:
Could you add a description about the interface change on this pull request for the record?

kwkbtr (approved):
Thank you!

Issue #169, opened 2023-07-12
Adam optimiser seems incompatible with Python 3.11 dataclasses


BogdanRajkov:
Importing `Adam` from `quri_parts.algo.optimizer` while using Python 3.11 currently produces this error:

```ValueError: mutable default <class 'numpy.ndarray'> for field m is not allowed: use default_factory```

However, running the same line in Python 3.10 produces no errors. The difference seems to be due to a change in the `dataclasses` library.
In Python 3.10 we have (under `Python310/Lib/dataclasses.py`, lines 811-814):
```
# For real fields, disallow mutable defaults for known types.
if f._field_type is _FIELD and isinstance(f.default, (list, dict, set)):
    raise ValueError(f'mutable default {type(f.default)} for field '
                     f'{f.name} is not allowed: use default_factory')
```

Compared with Python 3.11 in which the same snippet is: (under `Python311/Lib/dataclasses.py`, lines 811-816)
```
# For real fields, disallow mutable defaults.  Use unhashable as a proxy
# indicator for mutability.  Read the __hash__ attribute from the class,
# not the instance.
if f._field_type is _FIELD and f.default.__class__.__hash__ is None:
    raise ValueError(f'mutable default {type(f.default)} for field '
                     f'{f.name} is not allowed: use default_factory')
```

As you can see, the criterion for immutability has changed, hence the new error. This should be relatively easy to fix by using `default_factory`, as suggested by the error message. To be honest, this might be more of an issue with the Python 3.11 implementation than with `quri-parts`, seeing how it flags a read-only view of a `numpy` array as mutable, but there's not much we can do about that now.

kwkbtr:
Thank you for a detailed report!
I agree that what we can do right now is to use `default_factory` returning the `_const_zero_array`.
We would like to fix the issue but please feel free to create a pull request if you are interested.

BogdanRajkov:
I've opened a PR now, should be fixed!

kwkbtr:
Thank you for the PR!
We merged it and released a new version of QURI Parts (0.13.0).

PR #171, opened 2023-07-13
Replacing unhashable defaults with factories

BogdanRajkov:
Resolves issue #169

I've checked all instances of frozen dataclasses in the library and think they should all work now, at least I couldn't generate an error while using Python 3.11.

### DIFF
--- a/packages/algo/quri_parts/algo/optimizer/lbfgs.py
+++ b/packages/algo/quri_parts/algo/optimizer/lbfgs.py
@@ -9,7 +9,7 @@
 # limitations under the License.
 
 import warnings
-from dataclasses import dataclass, replace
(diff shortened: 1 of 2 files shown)


BogdanRajkov:

packages/algo/quri_parts/algo/optimizer/lbfgs.py
```diff
@@ -9,7 +9,7 @@
 # limitations under the License.
 
 import warnings
-from dataclasses import dataclass, replace
+from dataclasses import dataclass, replace, field
```
kwkbtr: Fix for isort:
```suggestion
from dataclasses import dataclass, field, replace
```

kwkbtr (approved):
Thank you!

PR #185, opened 2023-08-01
Kupccgsd identify parameter

dchung0741:
Identify particular parameters in kupgccsd ansatz

- (Fix): Non-Singlet excitation kUpGCCSD ansatz: parameter count reduced to $3N(N-1)/2$ where $N$ is the number of spatial orbitals.
- (Add): Singlet excitation kUpGCCSD ansatz: parameter count reduced to $N(N-1)$ where $N$ is the number of spatial orbitals.

### DIFF
--- a/packages/openfermion/quri_parts/openfermion/ansatz/kupccgsd.py
+++ b/packages/openfermion/quri_parts/openfermion/ansatz/kupccgsd.py
@@ -184,7 +184,7 @@ def get_independent_excitations(
             # if p^ q is in s_excs, q^ p is also in s_exc.
             continue
         independent_s_exc.append(s_exc)
-        independent_s_exc_name.append(f"s_{x}_{q}_{p}")
+        independent_s_exc_name.append(f"t1_{x}_{q}_{p}")
 
     independent_d_exc = []
     independent_d_exc_name = []
@@ -195,7 +195,7 @@ def get_independent_excitations(
             # if q^ p^ s r is in d_excs, s^ r^ q p is also in d_excs.
             continue
         independent_d_exc.append(d_exc)
-        independent_d_exc_name.append(f"d_{x}_{s}_{r}_{q}_{p}")
+        independent_d_exc_name.append(f"t2_{x}_{s}_{r}_{q}_{p}")
 
     return (
         independent_s_exc,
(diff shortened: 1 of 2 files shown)


kwkbtr:
Can you write some comments in the PR description if there are breaking changes? (e.g. number/order of parameters etc.)

(By the way, #181 is already merged but could you add comments about breaking changes there too?)

dchung0741:
I just added comments on this PR page and the PR page for the UCCSD

packages/openfermion/quri_parts/openfermion/ansatz/kupccgsd.py
```diff
@@ -55,53 +60,142 @@ def __init__(
         fermion_qubit_mapping: OpenFermionQubitMapping = jordan_wigner,
         trotter_number: int = 1,
         delta_sz: int = 0,
+        singlet_excitation: bool = False,
```
kwkbtr, line 68: Can you add docstring for `singlet_excitation`?

dchung0741: I just added the docstring for it

packages/openfermion/quri_parts/openfermion/ansatz/kupccgsd.py
```diff
@@ -50,6 +50,9 @@ class KUpCCGSD(ImmutableLinearMappedUnboundParametricQuantumCircuit):
           :class:`Operator`
         trotter_number: Number for first-order Trotter product formula.
         delta_sz: Changes of spin in the excitation.
+        singlet_excitation: If ``True``, single excitations that start
+            from and end at the same spatial orbitals will share the same
+            circuit parameter.
```
kwkbtr: I'm a bit confused: in my understanding "singlet excitation" and "single excitations" are different. Is this description correct?

dchung0741: In the context of kUpGCCSD this description is correct. Singlet excitation means that the excitation operator $T$ should satisfy $[T, S_x] = [T, S_y]= [T, S_z] = 0$. For UCCSD, this condition imposes constraint on both single and double excitation. For kUpGCCSD, the double excitation operator is paired, so it is singlet automatically. Thus the singlet condition only imposes constraint on the single excitation part in the way I describe docstring.

Sorry if the names are a bit confusing, should I describe it in detail in the docstring?

kwkbtr: I see, thanks! Mentioning the following point in the docstring may be helpful for readers:

> For kUpGCCSD, the double excitation operator is paired, so it is singlet automatically

dchung0741: I just added the comment to the docstring.

kwkbtr (approved):

PR #189, opened 2023-08-07
Add bit length

dchung0741:
Create a `bit_length` function that returns the bit length from python int as well as numpy int8, int16, int32, int64

### DIFF
--- a/packages/core/quri_parts/core/utils/bit.py
+++ b/packages/core/quri_parts/core/utils/bit.py
@@ -14,9 +14,6 @@ import numpy as np
 
 
 def bit_length(bits: Union[int, np.int8, np.int16, np.int32, np.int64]) -> int:
-    assert isinstance(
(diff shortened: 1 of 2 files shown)


packages/core/quri_parts/core/utils/bit.py
```diff
@@ -8,6 +8,14 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+from typing import Union
+
+import numpy as np
+
+
+def bit_length(bits: Union[int, np.int8, np.int16, np.int32, np.int64]) -> int:
+    return int(bits).bit_length()
```
kwkbtr: Shall we add a docstring? (Though it is rather trivial)

dchung0741: I have added the docstring for it.

kwkbtr (approved):
👍

PR #205, opened 2023-08-24
Limit execution time

dchung0741:
1. Add `single_job_max_execution_time` to `QiskitRuntimeSamplingBackend` to let the user limit the time used by a single `(circuit, n_shot)` pair.
2. Add `strict_time_limit` option so that jobs that cannot be limited by the Runtime backend (those whose time limit is < 300 seconds) will be rejected.
3. Automatically distribute and add `max_execution_time` option to `Runtime Sampler` according to user input `single_job_max_execution_time` and time left set by the `total_run_time` option.


kwkbtr (approved):
👍

PR #231, opened 2023-10-05
Reduce CNOT Transpiler

lqtmirage:
- Implemented a CircuitTranspiler to reduce number of CNOT gates by replacing gate sequence patterns.
- The circuit replace pattern is based on the article below introduced by @Duckamo 
  - https://arxiv.org/abs/2308.00667


toru4838 (approved):
🚀🚀🚀

PR #240, opened 2023-10-18
Bug fix, performance improvements (quri-parts-itensor)

toru4838:
Bug fix and performance improvements:

- Add a flag to ensure_itensor_loaded()
- Normalize MPS before executing estimator/sampler
- Add license terms
- Fix typos

### DIFF
--- a/packages/itensor/quri_parts/itensor/estimator.py
+++ b/packages/itensor/quri_parts/itensor/estimator.py
@@ -44,7 +44,6 @@ ITensorParametricStateT: TypeAlias = ParametricCircuitQuantumState
 def _estimate(
     operator: Estimatable, state: ITensorStateT, **kwargs: Any
 ) -> Estimate[complex]:
-    ensure_itensor_loaded()
     if operator == zero():
         return _Estimate(value=0.0, error=0.0)
     qubits = state.qubit_count
@@ -86,6 +85,7 @@ def create_itensor_mps_estimator(
         maxdim: The maximum numer of singular values.
         cutoff: Singular value truncation cutoff.
     """
+    ensure_itensor_loaded()
 
     def estimator(operator: Estimatable, state: ITensorStateT) -> Estimate[complex]:
         if maxdim is not None:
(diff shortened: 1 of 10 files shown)


packages/itensor/quri_parts/itensor/estimator.py
```diff
@@ -57,15 +66,19 @@ def _estimate(
...
-    exp: float = jl.expectation(psi, op)
 
     # See https://github.com/QunaSys/quri-parts/pull/203#discussion_r1329458816
     error = 0.0
     if "maxdim" in kwargs or "cutoff" in kwargs:
         error = np.nan
 
+    # calculate expectation value
+    if kwargs:
+        psi = jl.normalize(psi)
```
kwkbtr: What's the rationale of this `if kwargs:` condition?

kwkbtr: In my understanding, the state should always be normalized.
I guess it is the case that the state can be unnormalized when some options such as `maxdim` or `cutoff` are specified.
With that assumption:

* If calling `.normalize()` doesn't have performance penalty when the state is already normalized, then we can always call it without the `if` condition.
* If it is not the case, then I think the options (`maxdim`, `cutoff`, etc.) should explicitly be specified in the `if` condition.

toru4838: Thanks for the comments!

> In my understanding, the state should always be normalized.
> I guess it is the case that the state can be unnormalized when some options such as maxdim or cutoff are specified.

That's true.

> If calling .normalize() doesn't have performance penalty when the state is already normalized, then we can always call it without the if condition.

Every time you call `normalize()`, the norm (lognorm actually) of the MPS is computed regardless of its norm. So avoiding the call of `normalize()` when the MPS is normalized can be better.
https://github.com/ITensor/ITensors.jl/blob/e7740ebb8df26cbacc1671113656e702030e43bb/src/mps/abstractmps.jl#L1299-L1321

> If it is not the case, then I think the options (maxdim, cutoff, etc.) should explicitly be specified in the if condition.

I thought there might be an option which changes the norm of the MPS which I don't recognize. I just made a low risk choice (I should say I was a bit lazy 😅). I will change it to check if `maxdim`, `cutoff`, or `mindim` is specified.

kwkbtr (approved):
👍

PR #250, opened 2023-10-31
Fix overlap estimator bug

### LINKED ISSUE #249: Overlap estimator bug
This line sometimes throws a key error:
https://github.com/QunaSys/quri-parts/blob/407ae825f995c7db9afd0c2983c1932407ea4f29/packages/core/quri_parts/core/estimator/sampling/overlap_estimator.py#L37

This one too
https://github.com/QunaSys/quri-parts/blob/407ae825f995c7db9afd0c2983c1932407ea4f29/packages/core/quri_parts/core/estimator/sampling/overlap_estimator.py#L42

It seems kind of obvious in hindsight, but the `self._sampling_counts.keys()` should be checked before requesting a specific count.

ThomasenQunasys:
closes #249

### DIFF
--- a/packages/core/quri_parts/core/estimator/sampling/overlap_estimator.py
+++ b/packages/core/quri_parts/core/estimator/sampling/overlap_estimator.py
@@ -34,21 +34,15 @@ class _OverlapEstimate:
 
     @cached_property
     def value(self) -> float:
-        if 0 in self._sampling_counts.keys():
-            val = self._sampling_counts.get(0, 0) / self._total_counts
-        else:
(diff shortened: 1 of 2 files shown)


kwkbtr (commented):
There is no problem in this fix but it is a bit redundant; when you use `.get()`, you don't need to check the existence of the key with `.keys()`.

ThomasenQunasys:
Ah. I forgot to save the last change in my local working tree before committing.

kwkbtr (approved):

PR #255, opened 2023-11-08
Add dependency on qiskit-aer

toru4838:
`qiskit-aer` is used in `quri-parts-qiskit` package but it is not specified as a dependency in `pyproject.toml`.
This PR adds a dependency on `qiskit-aer` explicitly.


kwkbtr:
I thought Aer is only used in test cases. If so, it should be listed in dev dependencies.
Is Aer used in library code too?

toru4838:
Thanks for the comment.
It is also used in tutorials, but I will put it in dev dependencies 👍

kwkbtr (approved):

kwkbtr (approved):

PR #261, opened 2023-11-20
Update license files

toru4838:


dchung0741 (approved):
Thanks for the fix! 👍

PR #272, opened 2023-11-29
Check operator estimatable

dchung0741:
Currently in sampling estimate computations, there is no mechanism to check whether the qubit count of the operator is larger than the state on which it is estimated. A new function is_estimatable is added for checking that.

Note:
Qulacs estimators are not checked by the new function as Qulacs already performs the check by itself.


rykojima (approved):
LGTM!

PR #274, opened 2023-11-30
Execute Pauli(Rotation)DecomposeTranspiler before converting circuit (quri-parts-itensor)

toru4838:
In current implementation `QuantumCircuit` which contains `Pauli` or `PauliRotaion` gates is not available for `quri-parts-itensor`'s estimator and sampler.
This PR introduces `Pauli(Rotation)DecomposeTranspiler` as a default argument of `convert_circuit()` and it is executed before converting the circuit to ITensor's.


rykojima (approved):
LGTM!

PR #277, opened 2023-12-04
Add `__eq__()` method to `(Parametric)QuantumGate`

toru4838:
Comparing two `Pauli` or `PauliRotation` gates that have equivalent matrix representation but different order of `target_indice` and `pauli_ids` gives the result we don't expect.

For example,

```python
Pauli(target_indices=(0,1), pauli_ids=(1,2)) == Pauli(target_indices=(0,1), pauli_ids=(1,2))  # True
Pauli(target_indices=(0,1), pauli_ids=(1,2)) == Pauli(target_indices=(1,0), pauli_ids=(2,1))  # False
```

This PR adds `__eq__()` method to `QuantumGate` and `ParametricQuantumGate` that compares the pairs of `target_index` and `pauli_id`

### DIFF
--- a/packages/circuit/quri_parts/circuit/gate.py
+++ b/packages/circuit/quri_parts/circuit/gate.py
@@ -9,7 +9,9 @@
 # limitations under the License.
 
 from collections.abc import Sequence
-from typing import NamedTuple
+from typing import NamedTuple, TypeVar
+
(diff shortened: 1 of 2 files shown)


dchung0741 (approved):
LGTM!

PR #287, opened 2023-12-15
Base ActiveSpaceMolecularOrbital class

dchung0741:
Add base `ActiveSpaceMolecularOrbital` class for generic electron-integral-related usage.

Motivation of this PR:
The `to_active_space_mo_int` method of the Protocol class `AOeIntSet` takes in `ActiveSpaceMolecularOrbitals` originally, which is an object reserved for non-relativistic molecular systems. We want to follow the same interface for generic systems other than molecular systems, whose corresponding active space MO should not be a subclass of the active space MO for molecular systems. Thus, it makes sense to have a base class for active space MO and implement it for different systems.


toru4838 (approved):
🚀🚀🚀

PR #289, opened 2023-12-18
Default backend max shot

dchung0741:
Setting the `Qiskit(Runtime)SamplingBackend`'s default max shot of a single batch to 1 million shots.
The default setting is triggered and warning message is thrown if no `max_shot` setting can be found for the real device or Aer simulator.


toru4838 (approved):
LGTM!

PR #294, opened 2023-12-26
add overload signature

dchung0741:
- Add overload variants to quantum_state.
- Make arguments other than n_qubits in quantum_state as kwargs


kwkbtr (approved):
👍

PR #300, opened 2024-01-09
Add TketTranspiler for optimization

lqtmirage:
- Added a transpiler that uses tket's compilation capabilities to perform circuit transformation for a given backend, gate set conversion, circuit simplification, etc.
  - https://cqcl.github.io/pytket/manual/manual_compiler.html


toru4838 (approved):
👍

PR #303, opened 2024-01-18
fix empty circuit error

dchung0741:
Fix error when sampling with empty circuit.


rykojima (approved):
Thank you for your implementation! LGTM!

PR #304, opened 2024-01-19
Braket inv circuit converter

dchung0741:
- Inverse circuit converter from Braket to QURI Parts is supported.
    - gate_from_braket
    - circuit_from_braket
- SqrtX, SqrtXdag, U1, U2, U3 are supported with their counter parts in Braket.


rykojima (approved):
LGTM!

PR #306, opened 2024-01-23
Cache PauliLabel instances in a `weakref` dictionary

toru4838:
When generating numerous `Operators`, there is a considerable amount of duplicated `PauliLabel`s that get allocated in memory. This pull request introduces a caching mechanism (using `weakref`) to prevent unnecessary memory wastage.


dchung0741 (approved):
Thanks for the implementation. Looks good to me!

PR #308, opened 2024-01-29
Uccsd fix

dchung0741:
- Allow UCCSD ansatz with odd number of electrons when `singlet_excitation = False`
- Improve docstrings


toru4838 (approved):
Thank you!

PR #323, opened 2024-03-18
Bug Fix: sz=0 ignored in post selection filter

dchung0741:
When `sz=0`, it is currently recognized as `sz=None` and does not check `sz` for the state


toru4838 (approved):
👍👍👍

PR #338, opened 2024-05-31
Bug of draw_circuit function

ayakatoayaka:
I fixed a bug,

>  IndexError: index 1 is out of bounds for axis 0 with size 1

, when  we use `draw_circuit` method while there is no gate on the circuit. 
In a result, when there is no gate on the circuit, 'draw_circuit' shows lines(`-`)  representing the number of qubits.


ayakatoayaka:

toru4838 (approved):
Thank you!

PR #342, opened 2024-06-28
From qulacs converter2

HayatoYunoki:
make circuit converter to convert qulacs to quri-parts and add test file for it


toru4838 (approved):
Thank you!!

PR #345, opened 2024-07-04
Qiskit 1 update

dchung0741:
Update the `quri_parts.qiskit` package to support qiskit >= 1.0.

Breaking changes:
- the `quri_parts.qiskit.operator.operator_from_qiskit_op` function supports only `SparsePauliOp` input.

Other interfaces remain compatible with earlier versions of `quri_parts.qiskit`


toru4838 (approved):
Thank you!!

PR #346, opened 2024-07-08
Make ImmutableQuantumCircuit.depth a cached property

kwkbtr:
This brings a performance improvement when the `.depth` property is accessed repeatedly.


toru4838 (approved):
🚀🚀🚀

PR #347, opened 2024-07-08
Freeze the circuit before converting to a Qulacs circuit with a noise model

kwkbtr:
This will bring a performance improvement together with #346.


toru4838 (approved):
LGTM!

PR #348, opened 2024-07-09
fix licensed workflow

toru4838:


kwkbtr (approved):
I tried running the workflow and it detected the license issues, so this fix is working 👍 
https://github.com/QunaSys/quri-parts/actions/runs/9849993476/job/27194545542#step:8:12

PR #362, opened 2024-10-10
Merge QuantumCircuit backend written in Rust

yasuo-ozu:
In 2024, we will release a new quantum circuit backend. This will provide users with a faster experience in quantum circuit generation and transpilation. Circuit generation and transpilation are crucial use cases for QURI Parts. For example, in workflows where generating large ansatz circuits or creating numerous instances of quantum circuits while varying parameters is required, the speed of quantum circuits is of paramount importance. This update will offer significant benefits to QURI Parts users, further solidifying QURI Parts' performance advantage over other libraries.

Additionally, as the user base of QURI Parts expands, this update makes QURI Parts easier to use. We have restructured the class design, which was previously considered complex, to make it more intuitive. For example, QuantumCircuit, which was previously a completely separate class from ImmutableQuantumCircuit, is now a subclass of ImmutableQuantumCircuit. This is because QuantumCircuit supports all the operations of ImmutableQuantumCircuit, in addition to allowing updates to the circuit. We have also shortened some of the longer class names to make them easier to understand. These changes have been made carefully to ensure backward compatibility. While the old class names are now deprecated, they can still be used.

# Key features

## Natively implemented classes

- `quri_parts.circuit.{circuit, circuit_parametric, gate, gates, noise}`
- Support faster circuit construction, parameter mapping, transpilation, noise simulation
- API is (almost) all preserved

## Parametric transpiler support

- `ParametricTranspiler` class, which allows to apply transpiler on ParametricQuantumCircuit directly
- You can wrap any `CircuitTranspiler` with `ParametricCircuitTranspiler`
- Additional simple transpiler for parametric circuits

## Class structure update, class names modification

- Reconstructed class structure (with backword compatibility)
- Deprecate longer class names and provide shorter one


yasuo-ozu:
Move to #363

PR #368, opened 2024-10-24
Bug fix and add tests for CliffordConversionTranspiler

lqtmirage:
- Fixed a bug in `CliffordConversionTranspiler` that caused qubit indices to be reassigned since using the raw cache.
- Add test cases.


kwkbtr (approved):
👍

PR #370, opened 2024-10-24
Circuit cost estimator

kwkbtr:
* Introduce `QubitProperty`, `GateProperty` and `DeviceProperty` under `quri_parts.backend.device` to describe noise characteristics of qubits and gates, and properties of a device such as the native gate set, connectivity and the noise model.
* Introduce "circuit cost estimator" functions: you can evaluate latency and fidelity of a specified circuit on a specified device.
* Device definitions that can be used for the latency and fidelity evaluation
  * Typical NISQ superconducting device
  * Typical NISQ trapped ion device
  * STAR architecture for early-FTQC
  * Clifford+T architecture for FTQC


dchung0741 (approved):
LGTM!👍

PR #381, opened 2024-10-25
DeviceProperty: support transpiler only for analyze

kwkbtr:
Some DeviceProperty requires for circuit analysis a transpiler different from the original transpiler for the architecture.


lqtmirage (approved):
LGTM!

PR #394, opened 2024-12-19
Update license files

toru4838:
Fix failing license check https://github.com/QunaSys/quri-parts/actions/runs/12403727696


kwkbtr (approved):
👍

PR #397, opened 2025-01-16
Noise model for nisq devices

lqtmirage:
- Add `NoiseModel` to typical NISQ `DeviceProperty`s
  - Apply `DepolarizingNoise` for 1 and 2 qubit gates, `BitFlipNoise` for `Measurement`
- Currently, t1 / t2 is not even reflected in the idling error during circuit fidelity estimation, so it is not included in the noise model
  - Changed to display a warning when a value other than `None` (default value) is passed for t1 / t2 when creating a `DeviceProperty`


yoooopeeee (approved):
👍

PR #400, opened 2025-01-31
add a measurement instruction for all qubits when converting QURI Parts quantum circuits to OpenQASM3

snuffkin:
In `quri-parts-openqasm`, when converting QURI Parts to OpenQASM 3, the `measure` instruction was not added.
As a result, there were cases where the converted OpenQASM 3 could not be executed on a quantum computer due to the lack of measurements.
Referring to `quri-parts-qiskit`, which adds `measure_all` during conversion, I modified the conversion process to add `c = measure q;` when converting QURI Parts quantum circuits to OpenQASM 3.


ThomasenQunasys:
@snuffkin Thank you for the PR, I will test and review it today!

ThomasenQunasys (approved):
Thanks @snuffkin , I don't see any problem with this fix. LGTM

PR #402, opened 2025-02-27
Fix license check

kwkbtr:
- Update licensed to the latest version (5.x)
- Fix the license cache file for Jinja2 that was broken and the cause of licensed error
- Update several license files and add a few allowed licenses


toru4838 (approved):
Thank you!

PR #403, opened 2025-02-27
Convert Identity to RZ when braket backend device does not support i gate

lqtmirage:
- For Braket devices, modified to convert Identity gate to RZ gate on the QURI Parts side when i gate is not supported by the backend device


toru4838 (approved):
Thank you!

PR #406, opened 2025-03-07
Port qsub as a submodule

lqtmirage:
- Ported qsub as a submodule under the Python^3.10 dependency.
- Other module dependencies remain ^3.9.8.
- Support only basic std libraries.
- Update licenses for several dependent packages.


yasuo-ozu:
Modify https://github.com/QunaSys/quri-parts-internal/blob/internal-main/.github/workflows/package.yml#L176C11-L176C50

yasuo-ozu (approved):

PR #409, opened 2025-03-31
Qubit mapping transpiler considering 2 qubit gate errors of actual devices

lqtmirage:
### Overview

**quri_parts.backend.coupling_error_map.py**

- Using the information from the actual devices and considering the 2 qubit gate (such as CNOT) errors, user can check if the number of effectively connected qubits is larger than the number of coupled qubits in the circuit.
- Search optimal qubit path from coupling map with 2 qubit gate errors for HE ansatz like circuits.
- Provide optimal qubit mapping feature for HE ansatz like circuits as a transpiler.

**quri_parts.circuit.transpile.analyze.py**

- Prepare analyzers of the circuits to find the coupled qubits, gate type weighted depth, number of gates satisfying the given condition, etc.


yasuo-ozu:
See https://github.com/QunaSys/quri-parts-internal/pull/106

PR #411, opened 2025-04-03
Enhance DeviceProperty for NISQ devices

lqtmirage:
### Extended functionality needed on the QURI Parts side to use NISQ devices with QURI VM

**Added new foundation gate for Qiskit**
- Added ECR gate

**Changed gate name type**
- Relaxed type constraint from GateNameType to str in DeviceProperty to accommodate each backend foundation gate
- Accompanying change to specify each qubit number of the underlying gate when creating a typical NISQ device

**Improved customizability of transpiler specification and optional features**
- User-specified transpilers can now be used when creating DeviceProperty
- Gate set conversion transpilers can now choose not to perform post-conversion verification
  - Gates that are not to be converted can now be passed through bare

**Fixed problems where necessary changes in previous PRs were being overwritten, etc.**
- Convert to rz gates for devices that do not support i-gates in Braket
- Provide noise models for typical ion trap/superconducting devices
- Bug fix in conversion of Identity gate to Qulacs

### DIFF
--- a/packages/braket/quri_parts/braket/backend/transpiler.py
+++ b/packages/braket/quri_parts/braket/backend/transpiler.py
@@ -10,7 +10,12 @@
 
 from braket.aws import AwsDevice
 
-from quri_parts.circuit.transpile import CZ2CNOTHTranspiler, SequentialTranspiler
+from quri_parts.circuit.transpile import (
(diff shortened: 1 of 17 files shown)


toru4838 (approved):
Thank you!

PR #412, opened 2025-04-07
Bug fix in conversion of Pauli / PauliRotation gate to Qiskit

lqtmirage:
- Fixed a bug in the code that converts `Pauli` / `PauliRotation` gates directly to Qiskit (without disassembly).
  - Test cases did not cover the conversion code for individual gates, and it is expected that the test cases could not check the code that was no longer working due to Qiskit version changes, etc.
- Modified test cases so that the gate conversions are tested.


toru4838 (approved):
Thank you! 🚀🚀🚀

PR #431, opened 2025-05-07
Integrate with setuptools-rust-bundled

yasuo-ozu:
This PR modifies the installation process of its sdist package.

- It no longer uses rust toolchain on the system.
- We can support more OS / Arch, which we do not generate wheel package for.
- This change will fix the installation problems on some HPC environments, like Fugaku. (For that we formally should install Rust toolchain manually)

# Limitation

For now, we have to install `quri-parts-rust` package separately in workflows like:

```shell
poetry run pip install -vvv -e packages/rust
poetry install -vvv --only main,dev,doc
```

I don't know why, but it maybe compatibility problem with Poetry.


toru4838 (approved):
Thank you!

PR #442, opened 2025-06-12
add qpe.py

ThomasenQunasys:


packages/qsub/quri_parts/qsub/lib/qpe.py
```diff
@@ -0,0 +1,94 @@
...
+        pqs, sqs = qubits[:bits], qubits[bits:]
+
+        builder.add_op(LineH(bits), pqs)
+        for k in range(bits):
+            for _ in range(2**k):
+                builder.add_op(Controlled(op), (pqs[k], *sqs))
+        builder.add_op(QFTdag(bits), pqs)
+
+
+QPE, QPESub = param_opsub(_QPE)
```
dchung0741, line 88: Maybe adding some docstrings here like [this](https://github.com/QunaSys/qp/blob/main/plus/qsub/quri_parts/plus/qsub/lib/amplitude_amp/rotation.py#L55-L74) can be helpful for the users

dchung0741 (commented):
I think adding some docstrings above the ops can be helpful. 
Also, maybe we want to add some descriptions to the PR

ThomasenQunasys:
@dchung0741 I added a docstring that describes the QPE and QPEListUk ops with reference to Nielsen and Chuang

dchung0741 (approved):
thank you! LGTM

PR #445, opened 2025-06-27
Use limited api

yasuo-ozu:
We can build the native binaries for every supporting Python interpreters using PyLimitedAPI limitation, instead of building native binaries for each interpreters individual.

It saves compile time, maintainance cost


yasuo-ozu:
| Before (individual binaries for py9, 10, 11, 12) | After (same binary for each python ver) |
|-----|-----|
| image | image |

yasuo-ozu:
I added some tests, so we have more workflow jobs now

packages/circuit/tests/circuit/test_inverse.py
```diff
@@ -107,7 +107,8 @@ def _assert_inverse_gates(a: QuantumGate, b: QuantumGate) -> None:
 def _assert_inverse_circuits(a: QuantumCircuit, b: QuantumCircuit) -> None:
     if a != inverse_circuit(b):
         raise AssertionError(
-            f"a != inverse_circuit(b). {a=}, {b=}, {inverse_circuit(b)=}"
+            f"a != inverse_circuit(b). {a=}, {b=}, {inverse_circuit(b)=},"
```
ThomasenQunasys: I am just curious, but is this change related to the compiler issue?

ThomasenQunasys:
Thanks @yasuo-ozu 

The workflow seems fine to me, and I confirmed that the tests also run properly. One thing I read about limited api is that

> Running in the Limited API has a notable performance cost. If this is a concern then you should measure it for your own module. Some rough guidelines follow:
> 
> Where the majority of the work involves C-level code the performance loss is likely to be low. This includes code that makes heavy use of typed memoryviews, or code that mainly calls an external C library.
> 
> Where the majority of the work involves interacting with Python objects the cost is likely to be more significant. As an example, compiling the Cython compiler with the regular C API gives a ~35% speed-up compared to not compiling the Cython compiler. Compiling the Cython compiler in the Limited API gives a 0-10% speed-up (depending on the exact version used).
> 
> If you are prepared to restrict yourself to Python versions 3.12+, then Cython will use the “vectorcall” interface in Limited API mode. This doesn’t enable any new functionality, but it does give a noticeable performance improvement. (Outside of the Limited API, Cython almost always uses this interface).

from https://docs.cython.org/en/latest/src/userguide/limited_api.html. Have you tested if there is any noticable difference?

.github/workflows/package.yml
```diff
@@ -167,42 +252,102 @@ jobs:
...
+        system:
+          - os: "ubuntu-latest"
+            target: "x86_64-unknown-linux-gnu"
+          - os: "macos-13"
+            target: "x86_64-apple-darwin"
+          - os: "macos-latest"
+            target: "aarch64-apple-darwin"
+          - os: "windows-latest"
+            target: "x86_64-pc-windows-msvc"
+        python-version: [ "9", "10", "11", "12" ]
```
ThomasenQunasys, line 278: Is it fine to exclude "13"? if I understand correctly, limited api supports it too, and the original mention by tsutsui-san was about 3.13 https://qunasys.slack.com/archives/C02MBHDSHNU/p1754381216718859

yasuo-ozu: I'll add 13 support in next PR. This PR is for build changes.

yasuo-ozu:
> Have you tested if there is any noticable difference

No I not tested, but as I know it is only of FFI costs, that specially affects PyList and PyTuple.
in QuantumCircuit, it will affect on function calls (like QuantumCircuit.add_XXX_gate()). It is not fundamental, because the most time-consuming thing is Python interpreter.

It will also affect on UnitaryMatrix gate passing from / into Rust, but it is also heavy without PyLimitedApi, so we should consider another way (like using numpy, etc) to handle this.

ThomasenQunasys:
I see, thank you.

I am still waiting for your comment on https://github.com/QunaSys/quri-parts/pull/445#discussion_r2259575364, but I think it otherwise lgtm

packages/circuit/tests/circuit/test_inverse.py
yasuo-ozu:
```suggestion
            f"a != inverse_circuit(b). {a=}, {b=}, {inverse_circuit(b)=}"
```

ThomasenQunasys: I wasn't commenting on the added `,` on this line, rather, the fact that this file is being changed seems kind of random to me. Is it necessary for this PR or do you need to rebase this on main maybe?

ThomasenQunasys: I see this is the commit where you added it, https://github.com/QunaSys/quri-parts/pull/445/commits/a7ab320d5e40e7945c3582abb9a592cfd3c3de3c. I am not sure if you intend to keep it, since the commit is called `Add debug line`, but if it is intentional then there is no need to change it

ThomasenQunasys (approved):

PR #446, opened 2025-06-27
Fix inverse resolution and new Inverse Control resolver

dchung0741:
- Correct phase inversion
- Add inverse inverse resolver
- Correct resolution error of inverse controlled `Op`s by adding `register_controlled_resolver` function

Current issue: 
We can optimize controlled op if the op takes the form: $O = A^{\dagger} B A$. The optimization is that controlled-O can be implemented by $A^{\dagger} (c-B) A$ so that we don't need to control A. However, when an Op is subjected to this optimization, its inverse still controls the A, leading to inconsistent inverse control. The reason is that we only register how `c-O` is resolved, but whenever there is a `Inverse(Control(O))`, [this resolver](https://github.com/QunaSys/quri-parts/blob/main/packages/qsub/quri_parts/qsub/lib/std/inverse.py#L144) is applied first to change it into `Controlled(Inverse(O))`, then `Inverse(O)` is resolved into smaller ops and the `Controlled` is applied to them individually. The proper order is to apply the optimization resolver first and then invert all the ops.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -2,17 +2,17 @@ name: Test
 
 on:
   pull_request:
-    paths:
-      - packages/**/*.py
-      - packages/**/*.rs
-      - ./**/Cargo.toml
-      - ./**/Cargo.lock
-      - rust/**/*.rs
-      - poetry.lock
-      - .github/workflows/test.yml
-      - .github/actions/python-poetry-install/action.yml
-      - .github/actions/rust-install-cache/action.yml
-      - ./**/*.jl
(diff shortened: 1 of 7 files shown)


kwkbtr (commented):
LGTM in general.

* Since I merged another PR there is a conflict now. Can you merge or rebase on the main branch and resolve the conflict?
* A more informative pull request title would be better.

dchung0741:
I have updated the title and rebased onto main

kwkbtr (approved):

PR #466, opened 2025-09-19
Use common {mypy, flake8} settings across quri-{algo,vm,parts}, enable check on CI

yasuo-ozu:
Background: multiple `poetry.lock` per project makes separated duplicated venvs, and also we can't trust that all quri-* products are installable in one environment.

To minimize the diff, I downgraded `cirq-core` and `mypy` to adopt to the version that quri-parts used before.

Changes:
- CI files
- pyproject.toml
- poetry.lock (to change mypy and cirq-core version)
- quri-algo/ quri-vm/ (applying lint)


ThomasenQunasys:
@yasuo-ozu Thanks for this PR it will be nice to have. I think there are still quite a few things that need to be changed, but do you want me to review it already?

PR #473, opened 2025-10-16
Update 0_time_evo.ipynb

peggjt:
Added reminder on installation of quri-algo as this is not part of the default installation.


peggjt:

ThomasenQunasys:
Thank you James!
For getting started and installing quri-sdk (including quri-algo) we refer to https://quri-sdk.qunasys.com/quick_start, which is the starting point for users. So quri-algo actually is part of the default package and you don't need to explicitly install it. But I think that even if that was the case we probably assume that the users who are looking at this tutorial already have quri-algo installed.

PR #475, opened 2025-10-24
expanded the notebook on bell states.

peggjt:
I have expanded the notebook on bell states to include a description of bell states, their preparation, and interpretation of results. Note, results are in the z-basis. Further work can be done here to show results in the x-basis and how this impacts probability distributions.


ThomasenQunasys:
Thanks for this @peggjt , it looks like the rebase you did did not get rid of the commits we discussed from the internal repo. Could you try again? You may have to do it interactively using `git rebase -i`, then you will likely have to force push to your branch, be careful not to delete your own commits!

ThomasenQunasys (commented):
Hi @peggjt , it looks like you need to run `make clean-nb` before it will pass ci. Could you try running it? I don't have any further comments

ThomasenQunasys (approved):

PR #481, opened 2025-11-05
add interface

ThomasenQunasys:
Add
- state preparation interface
- adiabatic state preparation algorithm


ThomasenQunasys:
Since this was already approved internally (see https://github.com/QunaSys/quri-algo-internal/pull/75) I will merge it

PR #482, opened 2025-11-11
Fix sub collector cache

dchung0741:
Caching inside the `SubCollector` fails in the case when ArrayRef is used as an Op parameter. This is because ArrayRef is only the same as the other ArrayRef when they share the same memory address.

This PR fixes this issue by checking the which Ops in a Sub do not exist in the CallTable dictionary.


ThomasenQunasys (approved):
LGTM! 🚀

PR #497, opened 2025-12-17
Improve performance of add_UnitaryMatrix_gate #375

nils-wittemeier:
This addresses the performance issue raised in [#413](https://github.com/QunaSys/quri-sdk-issues/issues/413).

## Issue Summary
When running a simple QURI circuit that includes a large UnitaryMatrix gate, a **major performance degradation** was observed:

- Python 3.11 + quri_parts 0.16.0: ~0.3 seconds
- Python 3.12 + quri_parts (main/internal_main): ~50 seconds

That’s a slowdown of **more than 100×.**

## Solution
Speed up loops with Rayon.


quri-parts/packages/rust/src/circuit/gate.rs
```diff
@@ -251,6 +251,20 @@ pub struct GenericGateProperty {
     pub unitary_matrix: Option<Vec<Vec<Complex64>>>,
 }
 
+impl Default for GenericGateProperty {
```
yasuo-ozu: What is the raw impl of `Default` for ?

quri-parts/packages/rust/src/circuit/gate.rs
```diff
@@ -239,7 +239,7 @@ impl QuantumGate<Option<f64>> {
     }
 }
 
-#[derive(Clone, Debug, Default)]
+#[derive(Clone, Debug)]
```
yasuo-ozu:
```suggestion
#[derive(Clone, Debug, Default)]
```

yasuo-ozu (changes requested):

quri-parts/packages/rust/src/circuit/gate.rs
nils-wittemeier: Thank you for catching this. I forgot to remove the manual implementation.
I switched GenericGateProperty to #[derive(Clone, Debug, Default)] and dropped the manual impl.

yasuo-ozu (approved):

PR #499, opened 2025-12-18
Keep idle qubits after conversion

nils-wittemeier:
QURIPartsEvaluatorHooks.result() now sizes the resulting QuantumCircuit by the larger of:

- the highest gate index and
- the total number of allocated qubits.

This ensures idle qubits are preserved instead of being trimmed away.

This keeps the qubit count of the circuit consistent.

Addresses https://github.com/QunaSys/quri-sdk-issues/issues/484


toru4838 (approved):
👍

PR #496, opened 2025-12-17
Add hashing on QuantumCircuit/ImmutableQuantumCircuit

yasuo-ozu:

PR #500, opened 2025-12-25
Remove __hash__ implementation for QuantumCircuit, ParametricQuantumCircuit

yasuo-ozu:
ref: https://docs.python.org/3.14/reference/datamodel.html#object.__hash

Fix for #496 

remove quri-parts/poetry.lock


kwkbtr (commented):
`__eq__()` method should also be defined for `ImmutableQuantumCircuit`:

> If a class does not define an [`__eq__()`](https://docs.python.org/3.14/reference/datamodel.html#object.__eq__) method it should not define a [`__hash__()`](https://docs.python.org/3.14/reference/datamodel.html#object.__hash__) operation either

yasuo-ozu:
`eq` is implemented for ImmutableQuantumCircuit (see L10 in circuit.rs)

kwkbtr (approved):

PR #502, opened 2025-12-30
Add qubit-count validation when concatenating two circuits #389

nils-wittemeier:
Fixes https://github.com/QunaSys/quri-sdk-issues/issues/414. QuantumStateVector.with_gates_applied now surfaces the “different qubit numbers” error instead of the misleading TypeError when combining a 3‑qubit state circuit with a 4‑qubit gate circuit.

- Circuits now reject mismatched qubit counts up front inside `ImmutableQuantumCircuit.combine`, so `circuit + other` raises `ValueError` with a clearer explanation.
- `ImmutableQuantumCircuit.__add__`; propagates this exception
- `ImmutableQuantumCircuit.__add__`; still raises `NotImplementedError` for all other types of errors. This necessary because, in Python when `__add__(LHS,RHS)__` raises `NotImplemented`, the Python will try to call the RHS’s `__radd__`. (For more details see discussion on original PR)

### DIFF
--- a/quri-parts/packages/core/tests/core/state/test_state_helper.py
+++ b/quri-parts/packages/core/tests/core/state/test_state_helper.py
@@ -69,7 +69,7 @@ def test_quantum_state(
     state1 = quantum_state(n_qubits, bits=0b01)
     assert isinstance(state1, ComputationalBasisState)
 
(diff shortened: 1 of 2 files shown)


toru4838 (approved):
Thank you!

PR #511, opened 2026-03-04
optional openfermion, pyscf

ThomasenQunasys:


dchung0741 (commented):
@ThomasenQunasys Thanks for making this PR. 
I think it looks good and should be good to go after the CI failures are fixed

ThomasenQunasys:
> @ThomasenQunasys Thanks for making this PR. I think it looks good and should be good to go after the CI failures are fixed

Thanks! It seems like the workflows failed due to a connection time-out, it seems to pass now

dchung0741 (approved):
LGTM!

PR #515, opened 2026-03-24
Add total qubit count evaluator

Kazutaka333:
• add TotalQubitCountEvaluatorHooks to count # of data qubit + ancilla qubit

add tests for:
    • nested subcalls
    • repeated use of the same sub without overcounting

### DIFF
--- a/quri-parts/packages/qsub/quri_parts/qsub/eval/qubitcount.py
+++ b/quri-parts/packages/qsub/quri_parts/qsub/eval/qubitcount.py
@@ -58,3 +58,46 @@ class AuxQubitCountEvaluatorHooks(EvaluatorHooks[int]):
         call_stack: list[SubId],
     ) -> None:
         ...
+
+
+class TotalQubitCountEvaluatorHooks(EvaluatorHooks[int]):
+    def __init__(self) -> None:
+        self.reset()
+
+    def reset(self) -> None:
+        self._total_qubit_count = 0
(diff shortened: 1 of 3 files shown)


kwkbtr (commented):
The implementation looks good, but I feel it may be cleaner to reuse `AuxQubitCountEvaluatorHooks` implementation since the only difference is either adding the argument qubits of the entry MachineSub or not.

Kazutaka333:
> The implementation looks good, but I feel it may be cleaner to reuse `AuxQubitCountEvaluatorHooks` implementation since the only difference is either adding the argument qubits of the entry MachineSub or not.

I refactored TotalQubitCountEvaluatorHooks to subclass AuxQubitCountEvaluatorHooks and reuse its logic

quri-parts/packages/qsub/quri_parts/qsub/eval/qubitcount.py
```diff
@@ -40,6 +40,11 @@ def enter_sub(
     ) -> bool:
         return sub.sub_id not in self._cache
 
+    def _set_root_result(
+        self, sub: MachineSub, qubits: Sequence[Qubit], regs: Sequence[Register]
```
kwkbtr: I'm not sure `qubits` and `regs` arguments are necessary, but the overall implementation looks good.

kwkbtr (approved):

PR #517, opened 2026-03-31
Turn Quri VM into a namespace package

nils-wittemeier:
Turn `quri_vm` into a namespace package.

Addresses https://github.com/QunaSys/quri-sdk-issues/issues/478


toru4838 (approved):
Thanks!

PR #523, opened 2026-04-21
Check Qubits and Registers in resolve()

yasuo-ozu:
Status = finished

# What?

- @yasuo-ozu made some mistakes writing sub resolver in qsub, and these kinds of debug is hard.
- It checks the returning Sub to check the miss.

### DIFF
--- a/quri-parts/packages/qsub/quri_parts/qsub/resolve/resolve.py
+++ b/quri-parts/packages/qsub/quri_parts/qsub/resolve/resolve.py
@@ -149,15 +149,19 @@ def resolve_sub(op: Op, repository: SubRepository = default_repository()) -> Sub
     if resolver:
         result = resolver(op, repository)
         if result is not None:
-            expected_qubits = set(Qubit(i) for i in range(op.qubit_count))
             actual_qubits = set(result.qubits)
+            assert actual_qubits == set(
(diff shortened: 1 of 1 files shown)


kwkbtr (commented):
Could you add some information useful for debugging to the assertion error?

- `op(.id)`
- The expected vs. actual number of qubits/registers

quri-parts/packages/qsub/quri_parts/qsub/resolve/resolve.py
```diff
@@ -145,6 +147,18 @@ def with_override(self, addition: SubRepository) -> SubRepository:
...
     resolver = repository.find_resolver(op)
     if resolver:
-        return resolver(op, repository)
+        result = resolver(op, repository)
+        if result is not None:
+            expected_qubits = set(Qubit(i) for i in range(op.qubit_count))
+            actual_qubits = set(result.qubits)
+            assert (
+                actual_qubits == expected_qubits
+            ), f"While resolving op {op!r}: expected {op.qubit_count} qubits, but got {len(actual_qubits)}."
```
kwkbtr, line 158: The error message seems a bit inappropriate since the assertion can fail even if the number of qubits are equal.

kwkbtr (approved):

PR #527, opened 2026-04-27
Qsub CompositeSubRepository and override mechanism

dchung0741:
1. Introduce `CompositeSubRepository` for additional `SubResolver`s that overrides default `SubResolver`s in parent `SubRepository` during `resolve_sub`. This ensures new `SubResolver`s registered to the parent `SubRepository` is always present no matter when the `CompositeSubRepository` with additional `SubResolver`s is created.
2. Introduced `simulator_repository` that automatically resolves the multi-controlled gates into MC backend gates


kwkbtr (approved):
- Please change the PR title to a meaningful one.
- Let's squash-merge to get rid of a bit messy commit history. -> If you squash-merge this, you will need to rebase #528 branch, which may be a bit cumbersome.

LGTM!

PR #529, opened 2026-04-28
Add Rot op

yasuo-ozu:
This PR adds `Rot[n]` Op in qsub.

`Rot` is similar to `Shift` operation in classical computers, but it additionally move the MSB to LSB, that makes this operation unitary. `Rot` is traditional naming used in many classical ISAs.

### DIFF
--- a/quri-parts/packages/qsub/quri_parts/qsub/lib/std/swap.py
+++ b/quri-parts/packages/qsub/quri_parts/qsub/lib/std/swap.py
@@ -37,8 +37,8 @@ class _Rot(ParamUnitarySubDef[int]):
     position.
 
     The rotation is implemented as a ladder of adjacent :data:`SWAP`
-    operations, moving the state of the first qubit to the last qubit while
-    shifting every other qubit one position toward the first.
+    operations, moving the state of the first qubit to the last qubit
+    while shifting every other qubit one position toward the first.
     """
 
     name = "Rot"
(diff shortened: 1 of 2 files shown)


quri-parts/packages/qsub/quri_parts/qsub/lib/std/swap.py
```diff
@@ -9,8 +9,26 @@
...
+from quri_parts.qsub.opsub import ParamUnitarySubDef, param_opsub
 from quri_parts.qsub.register import DEFAULT_QNAME, QRegSpec
+from quri_parts.qsub.sub import SubBuilder
 
 from . import NS
 
 SWAP = Op(Ident(NS, "SWAP"), 2, self_inverse=True, qregs=(QRegSpec(DEFAULT_QNAME, 2),))
+
+
+class _Rot(ParamUnitarySubDef[int]):
```
Kazutaka333, line 35: can you add doc string for _Rot for user reference?

Kazutaka333 (changes requested):

quri-parts/packages/qsub/quri_parts/qsub/lib/std/swap.py
yasuo-ozu: done

Kazutaka333 (approved):
thank you!

PR #530, opened 2026-04-28
Implement __hash__() for Ident to support non-hashable Params like Op

yasuo-ozu:
This PR add support of non-hashable parameters in qsub

# Background

- There are some algorithms in qsub, which presumes that the `Op` is hashable (for example, `dict[Op, Sub]`)
- `Op` is currently unhashable because it has 

# Solution

In this PR, I add `Op.__hash__()` implementation manually, and hashing algorithm:

```py
def _param_hash(p: object) -> int:
    try:
        return hash(p)
    except TypeError as e:
        if isinstance(p, tuple):
            return hash(tuple(_param_hash(v) for v in p))
        elif isinstance(p, Op):
            return hash(p.id)
        elif hasattr(p, "base_id"):  # p is OpFactory
            return hash(p.base_id)
        else:
            raise e
```

This algorithm should be extended when we encounter some additional patterns that use other unhashable types in `Param`.

### DIFF
--- a/quri-parts/packages/qsub/quri_parts/qsub/op.py
+++ b/quri-parts/packages/qsub/quri_parts/qsub/op.py
@@ -45,7 +45,7 @@ def _param_hash(p: object) -> int:
             return hash(tuple(_param_hash(v) for v in p))
         elif isinstance(p, Op):
             return hash(p.id)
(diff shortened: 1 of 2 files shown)


yasuo-ozu:
#527 solves this problem

PR #539, opened 2026-06-03
Tensornetwork workspace

ThomasenQunasys:
# Tensornetwork Workspace

Convenience functions for TensorNetwork

## List of changes

- quri-parts/packages/tensornetwork/quri_parts/tensornetwork/circuit/__init__.py:
  - `TensorNetworkLayer` now has an `extend` method.
  - The logic to connect tensornetwork layers has been refactored
- quri-parts/packages/tensornetwork/quri_parts/tensornetwork/operator/__init__.py:
  - Added `contract_with` method which is now the main entry-point for observable estimation
  - This is more robust because it allows us to implement operator dependent contraction logic, instead of leaving it to the estimators. For MPOs this construction is better, especially depending on the contraction strategy used.
- quri-parts/packages/tensornetwork/quri_parts/tensornetwork/state/__init__.py:
  - Introduces `MultiMappedNode` for MPS and other uses
  - Contains convenience methods for detecting internal edges of MPOs and MPSs


ThomasenQunasys:
@dchung0741 this is the PR from yesterday, only difference is that I noticed that the branch contained a poetry.lock for quri-parts, which I think we no longer use, so I removed it. Otherwise it is the same.

dchung0741 (approved):
LGTM

PR #544, opened 2026-06-09
Require numpy 2 on Python==3.13 to fix Windows segfault

Kazutaka333:
## Problem
Release CI segfaults on Windows + Python 3.13. numpy 1.x has no cp313 wheel, so pip source-builds it with MinGW ("CRASHES ARE TO BE EXPECTED") and it segfaults on import — surfacing during scipy/braket import.

## Fix
Split deps by `python_version` marker so Py3.13 uses prebuilt numpy 2 wheels:
- `quri-parts-core`, `quri-parts-circuit`: `numpy>=2.1.0` on Py3.13, `<2.0` otherwise.
- `quri-parts-cirq`: `cirq-core>=1.6,<1.7` on Py3.13 (cirq-core<1.4 pins numpy<2.0), `<1.4` otherwise.

Markers are pinned to `python_version == '3.13'` (not `>= '3.13'`) so Py3.14 is unaffected.

Py<3.13 is unchanged → mypy 1.8.0 typecheck stays green, no source changes.

Verified via Windows CI probe, uv resolution (both Python branches), and built-wheel metadata.

## pyqret lock workaround
Dropped `platform_machine` markers and the intel-mac pyqret wheels from `quri-parts-qret` to speed up `poetry lock` (mirrors QunaSys/quri-sdk-internal#471). A `TODO` marks restoring them once we switch to uv.

## CI notes
- Dropped `macos-15-intel` (x86_64) from the matrix — builds timed out at ~6h. aarch64 `macos-latest` still covered.
- Temporarily added this branch to the workflow push triggers to run the release workflow here; **revert before merge**.

### DIFF
--- a/.github/workflows/quri-sdk-package.yml
+++ b/.github/workflows/quri-sdk-package.yml
@@ -2,7 +2,7 @@ name: QURI SDK Package and release
 
 on:
   push:
-    branches: [main, 'quri-parts-qret*', 'fix/numpy2-windows-py313']
+    branches: [main, 'quri-parts-qret*']
   release:
     types: [published]
 
--- a/.github/workflows/quri-sdk-package.yml
+++ b/.github/workflows/quri-sdk-package.yml
@@ -2,7 +2,7 @@ name: QURI SDK Package and release
 
 on:
   push:
-    branches: [main, 'quri-parts-qret*']
+    branches: [main, 'quri-parts-qret*', 'fix/numpy2-windows-py313']
   release:
     types: [published]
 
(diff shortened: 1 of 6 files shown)


Kazutaka333:
ran CI for QURI SDK Package and release to confirm all pass 

https://github.com/QunaSys/quri-sdk/actions/runs/27184808583?pr=544

Kazutaka333:
passing run https://github.com/QunaSys/quri-sdk/actions/runs/27317489800

Kazutaka333:
included this change to speed up `poetry lock`
https://github.com/QunaSys/quri-sdk-internal/pull/471

PR #545, opened 2026-06-11
Temporary workaround for speeding up poetry lock

kwkbtr:
- Drop Intel mac wheels for pyqret
- Drop platform_machine marker for pyqret


Kazutaka333 (approved):
lgtm

PR #549, opened 2026-06-17
Fix qsub visualization bug with Cbz and Label

nils-wittemeier:
-Added a small, typed visualization layer (`_visualization.py`) with gate/conditional models, helpers to map to qulacsvis, and a custom Matplotlib drawer that draws conditional shading, classical register wires/labels, measurement arrows, classical gate boxes, and control dots.
-Refactored `visualize.py` to use those models: map ops to `GateType`, build register indices, collect `Cbz`/`Label` into `ConditionalBlocks`, and keep drawing `Op`, `Sub`, and `MachineSub` with the new drawer.


Kazutaka333 (approved):
LGTM

PR #553, opened 2026-06-30
Add CliffordTSetTranspiler

yasuo-ozu:
## Summary

Adds `CliffordTSetTranspiler`, a `CircuitTranspiler` that transpiles an arbitrary `QuantumCircuit` into an equivalent one containing only Clifford+T gates (H, X, Y, Z, S, Sdag, SqrtX, SqrtXdag, SqrtY, SqrtYdag, CNOT, CZ, SWAP, T, Tdag).

It is implemented as a `SequentialTranspiler` composing existing transpilers:
- Decompose single-/two-qubit unitary matrices and Pauli / Pauli-rotation / Toffoli gates.
- Convert RX / RY / U1 / U2 / U3 rotations into RZ + SqrtX.
- Fuse rotations, then map RZ angles to named Clifford+T gates via `RZ2NamedTranspiler(allow_t_tdag=True)`, and eliminate identities.

## Notes

- RZ gates are **not** approximated into a sequence of Clifford+T gates. Only RZ gates whose angle corresponds to a Clifford+T gate (within `epsilon`) are converted; RZ gates with other angles remain as RZ in the output.


Kazutaka333 (approved):
lgtm

yasuo-ozu:
Refine needed

PR #559, opened 2026-07-08
add benchmarking amplitude amplification example notebook

ALS15204:
This notebook is designed to demonstrate how to benchmark amplitude amplification from end-to-end with QURI SDK.

I used IBM and IonQ as examples and surface-code-lattice-surgery-2D to demonstrate.


SU2MM:
3 comments on this notebook:

1. I think that the limitation of the resource  estimation that comes from using a simple oracle should be made somewhat clearer. The total cost is roughly
$$\text{Total Cost} = k \times \left( \text{Cost}(A S_0 A^\dagger) + \text{Cost}(S_f) \right)$$, where S_f cost is related to  the oracle which if we insert the optimal value of k gives
$$\text{Total Cost} \approx \frac{\pi}{4}\sqrt{N} \times \left( \mathcal{O}(n) + \mathcal{O}(C_f(n)) \right)$$
The current estimate essentially only contains the first part inside the parenthesis. So if we have a costly oracle the cost of defining reflection with respect to its unitary  would be multiplied by the $\sqrt{N}$ factor and lead to quite a lot of additional overhead. I think it is worth making this clear. 
2. In order to run the notebook I need to install the qret library using the following command:
pip install quri-parts-qret --find-links https://github.com/QunaSys/quration/releases/expanded_assets/v1.0.1-2
as written in https://github.com/QunaSys/quri-sdk/blob/main/quri-parts/packages/qret/pyproject.toml
This is different from the instructions in the notebook which says to just install it as pip install quri-parts[qret]
3. I think it is useful to include a demonstration that the Grover circuit created in QSUB actually gives the target state with the theoretical success probability when the circuit is sampled. I have added a section that does this in the latest git commit to the branch.

PR #562, opened 2026-07-14
[tensornetwork] Fix gate bug

ThomasenQunasys:
Fixes a bug where R_y and U_2 gates are incorrectly encoded. Also fixes a bug where the U gates are incorrectly constructed.


dchung0741 (approved):
LGTM

PR #575, opened 2026-08-24
Add module-level docstrings to quri-parts circuit modules

nils-wittemeier:


nils-wittemeier:
@Kazutaka333, CI is fixed now and merge conflicts are resolved. While doing that, I noticed that it would probably be cleaner/more universal to replace "Key classes / functions:"  with "Contents" what do you think? If you disagree, I'd be happy to revert it.

PR #579, opened 2026-09-02
Support Qiskit 2 on Python 3.10–3.13

yasuhito:
## Problem

`quri-parts-qiskit` currently depends on Qiskit 1.x on Python 3.13 and earlier, and on Qiskit 2.x on Python 3.14 and later. This prevents it from being installed alongside projects that require Qiskit 2 on Python 3.10–3.13.

One affected downstream project is [Tranqu](https://github.com/oqtopus-team/tranqu), a quantum circuit transpiler that depends on Qiskit 2. Tranqu has an existing [PR to add bidirectional QURI Parts conversion](https://github.com/oqtopus-team/tranqu/pull/32), but the current dependency constraints cannot be resolved on Python 3.13 and earlier.

The QURI Parts Qiskit integration is already tested with Qiskit 2 on Python 3.14, and Qiskit 1.x is no longer maintained.

## Changes

- Use `qiskit>=2.2,<3.0` on every supported Python version.
- Use `qiskit-ibm-runtime>=0.47,<0.48` on every supported Python version.
- Remove comments that associate Qiskit versions with Python versions.
- Remove Qiskit 1.x and `qiskit-ibm-runtime` 0.29.x from the lockfile.

## Validation

Ran the Qiskit package test suite with Qiskit 2.5.2 and `qiskit-ibm-runtime` 0.47.0:

- Python 3.10: 58 passed, 17 skipped
- Python 3.13: 58 passed, 17 skipped
- `uv lock --check`
- `flake8`

The existing CI matrix covers Python 3.10 through 3.14, so this change makes every supported Python version exercise Qiskit 2.

### DIFF
--- a/quri-parts/packages/qiskit/pyproject.toml
+++ b/quri-parts/packages/qiskit/pyproject.toml
@@ -21,12 +21,8 @@ dependencies = [
     "quri-parts-core",
     "pydantic>=2.0,<3.0",
     "networkx",
-    # qiskit 1.x pulls symengine, which has no cp314 wheel; qiskit 2.2+ drops it
-    # and supports 3.14. The code guards the removed BackendV1 so both work.
-    "qiskit>=1.3.0,<2.0; python_version < '3.14'",
-    "qiskit>=2.2.0,<3.0; python_version >= '3.14'",
-    "qiskit-ibm-runtime>=0.29.0,<0.30.0; python_version < '3.14'",
-    "qiskit-ibm-runtime>=0.47.0,<0.48.0; python_version >= '3.14'",
+    "qiskit>=2.2.0,<3.0",
+    "qiskit-ibm-runtime>=0.47.0,<0.48.0",
 ]
 
 [project.urls]
--- a/quri-parts/packages/qiskit/quri_parts/qiskit/backend/utils.py
+++ b/quri-parts/packages/qiskit/quri_parts/qiskit/backend/utils.py
@@ -14,7 +14,7 @@ from typing import Callable, Mapping, MutableMapping, Optional, Sequence
 from qiskit.providers.backend import Backend, BackendV2
 
 try:
-    # BackendV1 was removed in qiskit 2.0 (qiskit 1.x / Python <3.14).
+    # BackendV1 was removed in qiskit 2.0.
     from qiskit.providers.backend import BackendV1
 
     _SUPPORTED_BACKENDS: tuple[type, ...] = (BackendV1, BackendV2)
--- a/quri-parts/packages/qiskit/quri_parts/qiskit/backend/connectivity.py
+++ b/quri-parts/packages/qiskit/quri_parts/qiskit/backend/connectivity.py
@@ -19,8 +19,8 @@ if TYPE_CHECKING:
     from qiskit.providers import BackendV1
 
 try:
-    # BackendV1 was removed in qiskit 2.0 (only present on qiskit 1.x, i.e.
-    # Python <3.14). Fall back to None so the legacy branch is skipped.
+    # BackendV1 was removed in qiskit 2.0. Fall back to None so the legacy
+    # branch is skipped.
     from qiskit.providers import BackendV1 as _BackendV1
 except ImportError:
     _BackendV1 = None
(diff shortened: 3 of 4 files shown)


quri-parts/packages/qiskit/pyproject.toml
```diff
@@ -21,12 +21,8 @@ dependencies = [
...
     "pydantic>=2.0,<3.0",
     "networkx",
-    # qiskit 1.x pulls symengine, which has no cp314 wheel; qiskit 2.2+ drops it
-    # and supports 3.14. The code guards the removed BackendV1 so both work.
-    "qiskit>=1.3.0,<2.0; python_version < '3.14'",
-    "qiskit>=2.2.0,<3.0; python_version >= '3.14'",
-    "qiskit-ibm-runtime>=0.29.0,<0.30.0; python_version < '3.14'",
-    "qiskit-ibm-runtime>=0.47.0,<0.48.0; python_version >= '3.14'",
+    "qiskit>=2.2.0,<3.0",
+    "qiskit-ibm-runtime>=0.47.0,<0.48.0",
```
toru4838, line 25: This makes runtime 0.47 the only installable version, but [backend/primitive.py](https://github.com/yasuhito/quri-sdk/blob/fix/support-qiskit-2-on-python-3-10-to-3-13/quri-parts/packages/qiskit/quri_parts/qiskit/backend/primitive.py) still calls the legacy API: 
- `Session(service=..., backend=...)` (L209, L366)
- `Sampler(session=..., options=...)` (L367, L376)

As a result, `QiskitRuntimeSamplingBackend` stops working on every Python version.
Could you migrate these calls to `Session(backend=self._backend)` and  `Sampler(mode=session, options=...)` in this PR?

quri-parts/packages/qiskit/quri_parts/qiskit/backend/utils.py
```diff
@@ -14,7 +14,7 @@
 from qiskit.providers.backend import Backend, BackendV2
 
 try:
-    # BackendV1 was removed in qiskit 2.0 (qiskit 1.x / Python <3.14).
+    # BackendV1 was removed in qiskit 2.0.
```
toru4838, line 17: The existing `QiskitRuntimeSamplingBackend` tests are all skipped on `runtime >=0.30`, so the green CI does not cover that path.
It might be better to add at least one test against the new API, e.g. patching Sampler in `quri_parts.qiskit.backend.primitive` and asserting it is constructed with `mode=session`.

toru4838 (commented):
@yasuhito Thank you for the PR!
I left a few comments. Please take a look at them.

This is unrelated to your changes, but the qiskit extra in [quri-parts/pyproject.toml](https://github.com/QunaSys/quri-sdk/blob/main/quri-parts/pyproject.toml) still carries `python_version < '3.14'`, so `pip install quri-parts[qiskit]` on Python 3.14 silently installs nothing.
https://github.com/QunaSys/quri-sdk/blob/17623d3dfb462d663e201a2bebc56172f015d49b/quri-parts/pyproject.toml#L30
Since this PR removes the last Python-version-conditional qiskit constraints, it might be good timing to clean this up as well, by dropping the marker and re-running uv lock.

quri-parts/packages/qiskit/pyproject.toml
yasuhito: Fixed in af7b590. Both session paths now use `Session(backend=...)`, and the sampler is constructed with `Sampler(mode=session, options=...)` for qiskit-ibm-runtime 0.47.

quri-parts/packages/qiskit/quri_parts/qiskit/backend/utils.py
yasuhito: Added regression tests in af7b590 for both implicit and context-managed sessions. They patch `Sampler`/`Session` and assert that `Sampler` receives the session through `mode=`.

toru4838: The new tests look great, and covering both session paths is more than I asked for. Thank you for the quick fix!

toru4838 (approved):
I am merging this. Thank you for the contribution!

PR #592, opened 2026-09-14
[CI] fix and align the PyPI upload jobs in the release workflow

toru4838:
## Background

The v0.27.0 release published QURI Algo and QURI VM to PyPI, but the QURI Parts and QURI SDK uploads failed, so those packages are still on 0.26.4.

Failed run: https://github.com/QunaSys/quri-sdk/actions/runs/34798468974

The four release jobs were not consistent with each other. QURI Parts and QURI SDK installed `twine==6.1.0` with `packaging==24.2`, while QURI Algo and QURI VM installed twine unpinned.
After the build backend moved from poetry-core to hatchling in #570, the distributions declare Metadata-Version 2.5, and packaging 24.2 only knows up to 2.4.

## Changes
1. Unpin twine in the QURI Parts and QURI SDK release jobs, so all four jobs install it the same way.
2. Add `--skip-existing` to all four uploads. The jobs upload independently, so a partial failure leaves PyPI half updated and a retry would otherwise fail on the files that are already published. The tradeoff is that uploading an already published version now succeeds silently.

These steps only run on `release` events, so this PR cannot exercise them. After merging, v0.27.0 will be retaggedand the release republished, which runs the upload jobs. The already published QURI Algo and QURI VM files will be skipped by change 2.


templepmet (approved):
LGTM!

PR #593, opened 2026-09-14
Use PyPI pyqret to fix QRET package uploads

templepmet:
## Issue

N/A

## Changes

The v0.27.0 release fails when uploading `quri-parts-qret` because its wheel declares direct GitHub URL dependencies, which PyPI rejects.
Use `pyqret>=1.1.0,<2.0.0` from PyPI so the published wheel and sdist contain a normal version requirement and installation needs no external release feed.

Failed release: https://github.com/QunaSys/quri-sdk/actions/runs/34820106167

- Replace the ten direct URL requirements and remove `allow-direct-references`.
- Lock pyqret 1.1.0 from PyPI; all other resolved package versions and sources are unchanged.
- Remove the obsolete Quration `--find-links` arguments from package installation and license workflows.
- Update three test assertions for renamed pyqret 1.1.0 backend properties.

**Validation**

- QRET tests with pyqret 1.1.0 on Python 3.14/macOS 26: 21 passed.
- `uv lock --check --python 3.11` with uv 0.11.17: passed.
- `uv build quri-parts/packages/qret --no-sources --python 3.11`: wheel and sdist built successfully.
- Both distributions pass `twine check`; their metadata contains `pyqret>=1.1.0,<2.0.0` and no direct URL requirements.

**Compatibility**

The upstream pyqret 1.1.0 macOS wheels require macOS 26+, whereas the previous wheels targeted macOS 15+.
Source installation on older macOS versions has not been validated.

## Progress

- [x] Main implementation written
- [x] Functions as intended
- [x] Unit tests have been written
- [ ] readme.md has been updated


toru4838 (approved):
Thank you!

I pushed the same commits to a `release-pyqret-pypi` branch and opened #594 as a draft, because the package build and install matrix only runs when the head branch starts with `release-`. The two steps this PR changes (the `pip install` calls) live in that matrix.

The matrix is green: https://github.com/QunaSys/quri-sdk/actions/runs/34830587711

I think it confirms:
* Every package builds, including `quri-parts-qret` without `allow-direct-references`.
* `pip install` works without the `--find-links` argument, so pyqret 1.1.0 resolves from PyPI on Linux, macOS, and Windows for Python 3.10 to 3.14.
* Tests pass against the installed wheels on the whole matrix.

I also downloaded the `quri-parts-packages-python` artifact from that run and read the METADATA of the built wheel. It declares `pyqret<2.0.0,>=1.1.0` and contains no direct URL requirement, which is what PyPI rejected with 400 in the last release.

I am merging this PR, and will close #594.

